PR #2, opened 2022-01-11
Account tile

banazsinjary:
account/profile tile for collections page


src/tiles/grid/account-tile.ts
```diff
@@ -2,22 +2,140 @@ import { css, html, LitElement } from 'lit';
...
+            <div class="archivist-since">
+              <h3>Archivist Since</h3>
+            </div>
+            <div class="year-holder">
+              <h3>${this.model?.addeddate?.getFullYear()}</h3>
+            </div>
+          </div>
+          <div class="status-holder">
+            <div class="patron">${accountIcon}</div>
+            <div class="stat-icons">
```
jbuckner, line 38: Minor terminology thing, but for this class, I would probably make it singular `stat-icon`. It's a class that gets applied to a single icon vs a container where you have many `stat-icons`.

src/tiles/grid/account-tile.ts
```diff
@@ -2,22 +2,140 @@ import { css, html, LitElement } from 'lit';
...
+          </div>
+          <div class="credit-holder">
+            <div class="archivist-since">
+              <h3>Archivist Since</h3>
+            </div>
+            <div class="year-holder">
+              <h3>${this.model?.addeddate?.getFullYear()}</h3>
+            </div>
+          </div>
+          <div class="status-holder">
```
jbuckner, line 36: Since many of these classes are only being used for a single element in the component, you can use `id` instead of `class`. Classes are typically used when you want to apply a style to many elements. In pre-web component days you had to be careful with `id` since every `id` on the page is supposed to be unique, but in web components, the ids only have to be unique within the component.

src/tiles/grid/account-tile.ts
```diff
@@ -2,22 +2,140 @@ import { css, html, LitElement } from 'lit';
...
+        font-weight: bold;
+        color: #2c2c2c;
+        margin: 0px;
+      }
+
+      .outter-holder {
+        background-color: #fcf5e6;
+        border: 1px #2c2c2c;
+        border-radius: 4px;
+        box-shadow: 1 1 2 0;
```
jbuckner, line 79: You need units on your non-zero numbers.

src/tiles/grid/icons/account.svg
```diff
@@ -0,0 +1 @@
+<svg viewBox="0 0 100 100" xmlns="http://www.w3.org/2000/svg"><path d="m89.6854559 79.6500588c1.7300364 6.4823648 2.180423 13.3122689 3.3145441 20.3499412h-86c.5683151-15.8558542 2.98334063-30.7849367 15.1676149-41.6581341 22.9948067-20.518674 59.250299-9.0032844 67.517841 21.3081929zm-40.0998307-79.6500588c10.872402.0493248 19.9700408 9.25722341 19.917959 20.1421788-.0829413 11.042868-8.9616237 19.8492523-20.0602807 19.8578212-11.1181198 0-19.9397193-8.7904706-19.9397193-19.8908727-.0327543-11.11998815 9.0125781-20.17487063 20.082041-20.1091273z" fill="#333" fill-rule="evenodd"/></svg>
```
jbuckner, line 1: You can remove the `svg` versions since we have the `ts` version.

src/tiles/grid/icons/account.ts
```diff
@@ -0,0 +1,11 @@
+import { html } from 'lit';
+
+export default html`
+  <svg viewBox="0 0 100 100" xmlns="http://www.w3.org/2000/svg">
```
jbuckner, line 4: Could you add a [`<title>`](https://developer.mozilla.org/en-US/docs/Web/SVG/Element/title) tag inside the svg files? It's how you add screen-reader accessible info to an svg image. There's also the [`<desc>`](https://developer.mozilla.org/en-US/docs/Web/SVG/Element/desc) but that's mainly for more complex images.

src/tiles/grid/account-tile.ts
```diff
@@ -2,22 +2,140 @@ import { css, html, LitElement } from 'lit';
...
+            <h1>${this.model?.identifier}</h1>
+          </div>
+          <div class="avatar-holder">
+            <div
+              class="avatar"
+              style="background: url('https://archive.org/services/img/${this
+                .model?.identifier}'); background-position: center;"
+            ></div>
+          </div>
+          <div class="credit-holder">
```
jbuckner, line 28: What does `credit` mean in this case?

src/tiles/grid/account-tile.ts
```diff
@@ -2,22 +2,140 @@ import { css, html, LitElement } from 'lit';
...
+        height: 40px;
+        margin-bottom: 5px;
+      }
+
+      .avatar-holder {
+        margin-bottom: 5px;
+        display: flex;
+        align-items: center;
+        justify-content: center;
+        max-height: 160px;
```
jbuckner, line 105: Do you need this `max-height` since the avatar image has a fixed 160x160? I think if we changed the cell height, we'd want this section to grow to take up the available height.

src/tiles/grid/account-tile.ts
```diff
@@ -2,22 +2,140 @@ import { css, html, LitElement } from 'lit';
...
+          <div class="credit-holder">
+            <div class="archivist-since">
+              <h3>Archivist Since</h3>
+            </div>
+            <div class="year-holder">
+              <h3>${this.model?.addeddate?.getFullYear()}</h3>
+            </div>
+          </div>
+          <div class="status-holder">
+            <div class="patron">${accountIcon}</div>
```
jbuckner, line 37: I would make these classes more descriptive, like `patron-icon` (could also become an `id`). `patron` sounds like the user themselves.

src/collection-browser.ts
```diff
@@ -16,6 +16,7 @@ import '@internetarchive/infinite-scroller';
 import './tiles/tile-dispatcher';
 import './tiles/loading-tile';
 import './sort-filter-bar/sort-filter-bar';
+import './tiles/grid/account-tile';
```
jbuckner, line 19: This should already be imported in `tile-dispatcher`.

src/tiles/grid/account-tile.ts
jbuckner, line 79: You need the unit of measurement on any non-zero values.

src/tiles/grid/account-tile.ts
```diff
@@ -2,22 +2,140 @@ import { css, html, LitElement } from 'lit';
...
     return css`
       h1 {
-        color: green;
-        margin-top: 0;
+        color: black;
+        font-size: 16px;
+        margin: 0;
+      }
+
+      h2 {
```
jbuckner, line 64: Are you using `h2` anywhere?

jbuckner (changes requested):
This is looking really good! You've used some nice layout techniques to get everything where it's supposed to be! A few comments and suggestions, but really nice progress 👏

src/tiles/grid/account-tile.ts
banazsinjary: This made me realize the css block was labeled "stats-icon" so the size wasn't being applied, lol fixed!

src/tiles/grid/account-tile.ts
```diff
@@ -2,22 +2,141 @@ import { css, html, LitElement } from 'lit';
...
+
+      .outer-holder {
+        background-color: #fcf5e6;
+        border: 1px #2c2c2c;
+        border-radius: 4px;
+        box-shadow: 1px 1px 2px 0px;
+        height: 100%;
+        display: flex;
+        align-items: center;
+        justify-content: center;
```
jbuckner, line 81: Suggestion: Remove `align-items` and `justify-content` and update the `inner-holder` to use padding. This approach is causing the `inner-holder` to float inside the `outer-holder` and we want the `inner-holder` to push up against the `outer-holder`.

src/tiles/grid/account-tile.ts
```diff
@@ -2,22 +2,141 @@ import { css, html, LitElement } from 'lit';
...
+        text-align: center;
+        width: 100%;
+      }
+
+      .inner-holder {
+        width: calc(var(--collectionBrowserCellWidth) - 10px);
+        height: 96%;
```
jbuckner, line 88:
```suggestion
        width: 100%;
        padding: 5px;
```
As we discussed, using a percentage as the height causes the margins to change as the cell grows and this is supposed to have 5px of padding. Setting the padding also allows you to not have to use the `calc()`.

src/tiles/grid/account-tile.ts
```diff
@@ -2,22 +2,141 @@ import { css, html, LitElement } from 'lit';
...
+        display: flex;
+        flex-direction: column;
+      }
+
+      #header-holder {
+        flex: 1 1 0%;
```
jbuckner, line 94: You should be able to just do the shorthand `flex: 1;` here.

jbuckner (changes requested):

jbuckner (approved):
Looks good! :shipit:

PR #4, opened 2022-02-01
WEBDEV-5018: Restyled collection tile

pezvi:
Updated to match new design

### DIFF
--- a/src/tiles/grid/collection-tile.ts
+++ b/src/tiles/grid/collection-tile.ts
@@ -11,28 +11,22 @@ export class CollectionTile extends LitElement {
 
   render() {
     return html`
-      <div
-        id="container"
-        @mouseover=${() => {
-          console.debug('mouseover');
-        }}
-        @focus=${() => {
-          console.debug('focus');
-        }}
-      >
+      <div id="container">
         <div id="collection-image-title">
-          <div
-            id="collection-image"
-            style="background-image:url('https://archive.org/services/img/${this
-              .model?.identifier}')"
-          ></div>
           <div id="collection-title">${this.model?.title}</div>
+          <div id="collection-image-container">
+            <div
+              id="collection-image"
+              style="background-image:url('https://archive.org/services/img/${this
+                .model?.identifier}')"
+            ></div>
+          </div>
         </div>
         <div id="item-count-container">
           <div id="item-count-image-container">${collectionIcon}</div>
           <div id="item-count-stacked-text">
             <div id="item-count">${this.model?.itemCount.toLocaleString()}</div>
-            <div id="items-text">${msg('ITEMS')}</div>
+            <div id="items-text">${msg('items')}</div>
           </div>
         </div>
       </div>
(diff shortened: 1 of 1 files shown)


src/tiles/grid/collection-tile.ts
```diff
@@ -21,18 +21,20 @@ export class CollectionTile extends LitElement {
         }}
```
jbuckner, line 21: Let's remove these `@mouseover` and `@focus` listeners.. I think I put them in trying to debug that Safari issue.

src/tiles/grid/collection-tile.ts
```diff
@@ -43,27 +45,38 @@ export class CollectionTile extends LitElement {
     const cornerRadiusCss = css`var(--collectionTileCornerRadius, 4px)`;
 
     return css`
+      #collection-image-container {
+        /* width: 18rem; */
+        height: 18.5rem;
```
jbuckner, line 50: I think you can remove this height and width with the flex updates suggested below.

src/tiles/grid/collection-tile.ts
```diff
@@ -43,27 +45,38 @@ export class CollectionTile extends LitElement {
...
+        display: flex;
+        justify-content: center;
```
jbuckner, line 52: I fiddled with the layout a bit to allow the tile to grow to its container size (currently it's fixed size) and allows the image to stay centered in the cell if the container changes sizes. Just a few `flex` suggestions:

```suggestion
        justify-content: center;
        align-items: center;
        flex: 1;
```

Allows us to resize the cell and your tile will always fit properly:

Screen Shot 2022-02-03 at 10 33 55

src/tiles/grid/collection-tile.ts
```diff
@@ -43,27 +45,38 @@ export class CollectionTile extends LitElement {
...
         background-size: cover;
       }
 
       #collection-image-title {
         background-color: #666;
-        padding: 1.5rem;
+        border: 1px solid #2c2c2c;
+        padding: 0.5rem;
         border-top-left-radius: ${cornerRadiusCss};
         border-top-right-radius: ${cornerRadiusCss};
```
jbuckner, line 71: Continuing from above:

```suggestion
        border-top-right-radius: ${cornerRadiusCss};
        display: flex;
        flex-direction: column;
        flex: 1;
```

src/tiles/grid/collection-tile.ts
```diff
@@ -73,37 +86,53 @@ export class CollectionTile extends LitElement {
         height: 4rem;
       }
 
+      #container {
+        box-shadow: 1px 1px 2px 0px;
+        border-radius: ${cornerRadiusCss};
```
jbuckner, line 91: Last one:

```suggestion
        border-radius: ${cornerRadiusCss};
        height: 100%;
        display: flex;
        flex-direction: column;
```

jbuckner (changes requested):
The tile is looking really good! A few suggestions, but it's really close!

src/tiles/grid/collection-tile.ts
pezvi: I omitted align-items:center. The collection image is intentionally aligned near the top (to match up with the alignment of other tiles).

jbuckner (approved):
Nice job! Looks good! 👍

PR #5, opened 2022-02-09
Add creator field to query

mc2:
List views require the creator field

Added to query.


mc2:
@jbuckner: fixed PR to commit into `initial_work`

mc2:
Fixed the conflict, rather than rebase. 

There have been commits in `list-compact` branch since that commit.

jbuckner (approved):
Looks good 👍

PR #6, opened 2022-03-04
Item tile component

dualcnhq:
Screen Shot 2022-04-08 at 08 56 56


src/tiles/grid/mediatype-icons.ts
```diff
@@ -0,0 +1,39 @@
...
+import { etreeIcon } from '../../assets/img/icons/mediatype/etree';
+import { imagesIcon } from '../../assets/img/icons/mediatype/images';
+import { filmIcon } from '../../assets/img/icons/mediatype/film';
+import { softwareIcon } from '../../assets/img/icons/mediatype/software';
+import { textsIcon } from '../../assets/img/icons/mediatype/texts';
+import { tvIcon } from '../../assets/img/icons/mediatype/tv';
+import { videoIcon } from '../../assets/img/icons/mediatype/video';
+import { webIcon } from '../../assets/img/icons/mediatype/web';
+
+@customElement('mediatype-icons')
```
jbuckner, line 14: I would probably make this singular since you're only displaying a single icon from it.

dualcnhq: this is moved in https://github.com/internetarchive/iaux-collection-browser/pull/8 - for final review for implementation here

src/tiles/grid/item-tile.ts
```diff
@@ -1,23 +1,323 @@
...
+          id="item-image"
+          style="background-image:url(${imgSrcUrl})"
+        ></div>`
+      : html`<div id="item-image-box">
+          <div
+            id="item-image-deemphasize"
+            style="background-image:url(${imgSrcUrl})"
+          ></div>
+          <div class="tile-action no-preview">Content may be inappropriate</div>
+        </div>`;
```
jbuckner, line 32: Suggestion: Since these two blocks have a similar purpose / layout, can we combine their DOM and just add the "deemphasize" messaging and blur when appropriate instead of two separate image divs?

src/tiles/grid/item-tile.ts
```diff
@@ -1,23 +1,323 @@
...
+            <span id="text-by">By:</span>
+            <span>${itemCreator}</span>
+          </div>
+        </div>
+        <div id="item-stats-container">
+          <div id="stats-holder">
+            <div class="col">
+              <mediatype-icon mediatype="${
+                this.model?.mediatype
+              }" showText="true">
```
jbuckner, line 68: In lit, when you're binding values from a property like this, it's typically best practice to use lit [bindings](https://lit.dev/docs/templates/expressions/) (put the dot or question mark before the attribute, depending on the type). This allows that property to be updated if the source value changes. ie:

```suggestion
              <mediatype-icon .mediatype=${
                this.model?.mediatype
              } ?showText=${true}>
```

jbuckner (approved):
Looks good! Let's merge it!

PR #8, opened 2022-03-09
Mediatype icons selection component

dualcnhq:
**Usage:**

```
import '../../mediatype-icon'; // path

<mediatype-icon mediatype="audio">
```


src/mediatype-icons.ts
```diff
@@ -0,0 +1,39 @@
...
+import { etreeIcon } from './assets/img/icons/mediatype/etree';
+import { imagesIcon } from './assets/img/icons/mediatype/images';
+import { filmIcon } from './assets/img/icons/mediatype/film';
+import { softwareIcon } from './assets/img/icons/mediatype/software';
+import { textsIcon } from './assets/img/icons/mediatype/texts';
+import { tvIcon } from './assets/img/icons/mediatype/tv';
+import { videoIcon } from './assets/img/icons/mediatype/video';
+import { webIcon } from './assets/img/icons/mediatype/web';
+
+@customElement('mediatype-icons')
```
jbuckner, line 14: Suggestion: Make this singular `mediatype-icon` since its only showing a single icon

src/mediatype-icons.ts
```diff
@@ -0,0 +1,39 @@
...
+      tv: tvIcon,
+      video: videoIcon,
+      web: webIcon,
+    };
+
+    return html`${mediaTypeIcons[this.mediatype]}`;
+  }
+
+  render() {
+    return html`${this.mediaTypeIcon}`;
```
jbuckner, line 37: Instead of two levels of `html` (here and on line 33), you can make `mediaTypeIcons` a private class property and  do `html ${this.mediaTypeIcons[this.mediatype]}` here

jbuckner (changes requested):

src/mediatype-icons.ts
```diff
@@ -11,29 +11,25 @@ import { tvIcon } from './assets/img/icons/mediatype/tv';
 import { videoIcon } from './assets/img/icons/mediatype/video';
 import { webIcon } from './assets/img/icons/mediatype/web';
 
-@customElement('mediatype-icons')
+@customElement('mediatype-icon')
 export class MediaTypeIcons extends LitElement {
```
jbuckner, line 15: Could you update the class name and file name to be singular as well?

src/mediatype-icons.ts
```diff
@@ -11,29 +11,25 @@ import { tvIcon } from './assets/img/icons/mediatype/tv';
...
-      software: softwareIcon,
-      texts: textsIcon,
-      tv: tvIcon,
-      video: videoIcon,
-      web: webIcon,
-    };
-
-    return html`${mediaTypeIcons[this.mediatype]}`;
-  }
+  @property() mediaTypeIcons: { [key: string]: any } = {
```
jbuckner, line 18: You don'y need to use `@property()` here. You only need that if the values are dynamic. You can just use `private`. And since this shouldn't be modified, you can notate that with `readonly`
```suggestion
  private readonly mediaTypeIcons: { [key: string]: any } = {
```

jbuckner (changes requested):

jbuckner (approved):
Thanks!

src/mediatype-icon.ts
```diff
@@ -0,0 +1,35 @@
...
+import { textsIcon } from './assets/img/icons/mediatype/texts';
+import { tvIcon } from './assets/img/icons/mediatype/tv';
+import { videoIcon } from './assets/img/icons/mediatype/video';
+import { webIcon } from './assets/img/icons/mediatype/web';
+
+@customElement('mediatype-icon')
+export class MediaTypeIcon extends LitElement {
+  @property({ type: String }) mediatype = '';
+
+  private readonly mediaTypeIcons: { [key: string]: any } = {
```
mc2, line 18: suggestion: 

`MediaTypeIcon` =>  `MediatypeIcon`
`mediaTypeIcons` => `mediatypeIcons`

I think our use of the term `mediatype` as a "unit" term. Like `software` (not `softWare`).

Also:  the lit file name convention for a class is to have a `-` between each camelCased term (don't remember where I saw this, but could look it up).
So, `mediatype-icon.ts` => `class MediatypeIcon`
Otherwise,
`media-type-icon.ts` => `class MediaTypeIcon`

mc2: The above lit file-name-matching-a-classname spec should probably be a blocker.

dualcnhq: updated PR, please check

src/tiles/grid/mediatype-icons.ts
```diff
@@ -0,0 +1,71 @@
...
+import { imagesIcon } from '../../assets/img/icons/mediatype/images';
+import { filmIcon } from '../../assets/img/icons/mediatype/film';
+import { softwareIcon } from '../../assets/img/icons/mediatype/software';
+import { textsIcon } from '../../assets/img/icons/mediatype/texts';
+import { tvIcon } from '../../assets/img/icons/mediatype/tv';
+import { videoIcon } from '../../assets/img/icons/mediatype/video';
+import { webIcon } from '../../assets/img/icons/mediatype/web';
+
+@customElement('mediatype-icon')
+export class MediatypeIcon extends LitElement {
```
jbuckner, line 15: Q: Is this a duplicate of the `MediatypeIcon` above?

dualcnhq: yes, will refactor this when we merge the other component

jbuckner (approved):

PR #9, opened 2022-03-09
Add format count, date utilities, i8n

mc2:
Created utils/formatDate and formatCount for proper formatting of dates and counts given format options. Unit tests.

WEBDEV-5177

### DIFF
--- a/src/utils/format-count.ts
+++ b/src/utils/format-count.ts
@@ -1,0 +1,94 @@
+import { msg, str } from '@lit/localize';
+
+export type NumberFormat = 'short' | 'long';
+type LabelFormat = 'short' | 'long';
+type Divisor = 1_000_000_000 | 1_000_000 | 1_000 | 1;
+
+/**
+ * Return the magnitude of a number.
+ * @param {number} number
+ * @param {NumberFormat} numberFormat
+ * @returns {Divisor}
+ */
+function magnitude(number: number, numberFormat: NumberFormat): Divisor {
+  let divisor: Divisor = 1;
+  if (number >= 1_000_000_000) {
+    divisor = 1_000_000_000;
+  } else if (number >= 1_000_000) {
+    divisor = 1_000_000;
+  } else if (number >= 1_000 && numberFormat === 'short') {
+    divisor = 1_000;
+  }
+  return divisor;
+}
+
+/**
+ * Round a number to the nearest magnitude.
+ * @param {number} number
+ * @param {Divisor} divisor
+ * @returns {number}
+ */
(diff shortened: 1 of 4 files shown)


dualcnhq:
can you give an example how to use this? - import then code snippet in a component?

dualcnhq (approved):

dualcnhq:
code looks good though need to fix the conflict

src/utils/format-date.ts
```diff
@@ -0,0 +1,22 @@
+export type DateFormat = 'short' | 'long';
+
+export function formatDate(
+  date: Date | undefined,
+  format: DateFormat = 'short',
+  locale: string = 'en-US'
+): string {
+  if (!date) return '';
```
jbuckner, line 8: I haven't seen where this is being consumed, but would it make sense to have this return `undefined` if the `date` is `undefined`?

mc2: @jbuckner Since the output of this method is used for UI, in `render()`, I'm guessing that the empty string is the best result of a date that can't be formatted.

jbuckner: Yeah potentially. That's why I was curious to see how it was being used, but this is fine 👍

mc2:
@dualcnhq 

```
import { formatCount } from '/src/utils/format-count';
...
// render() {
   <div>${formatCount(this.model?someNumber, 'short', 'short')}</div>

File: src/utils/format-date.ts [hunk 2]
```diff
@@ -0,0 +1,36 @@
...
+
+export function formatDate(
+  date: Date | undefined,
+  format: DateFormat = 'short',
+  locale: string = 'en-US'
+): string {
+  // Return blank if undefined
+  if (!date) return '';
+
+  const options: any = {
```
jbuckner, line 17: You can use `Intl.DateTimeFormatOptions` instead of `any` so you get type safety.

src/utils/format-count.ts
```diff
@@ -0,0 +1,95 @@
...
+ */
+export function formatCount(
+  count: number | undefined,
+  numberFormat: NumberFormat = 'long',
+  labelFormat: LabelFormat = 'short',
+  locale: string = 'en-US'
+): string {
+  // Return blank if undefined
+  const number = count ?? -1;
+  if (number < 0) {
```
jbuckner, line 89: Does this method not work with negative numbers?

jbuckner (approved):
One question and a suggestion, but looks good overall

src/utils/format-date.ts
mc2: Done

PR #10, opened 2022-03-17
List tiles compact and detailed

mc2:
Getting there


src/mediatype-icon.ts
```diff
@@ -62,7 +70,7 @@ export class MediatypeIcon extends LitElement {
         text-align: center;
       }
 
-      svg {
+      #icon svg {
```
jbuckner, line 73: What is the svg size behavior when it's not wrapped in the `icon` div?

src/tiles/tile-dispatcher.ts
```diff
@@ -5,6 +5,7 @@ import {
   SharedResizeObserverInterface,
   SharedResizeObserverResizeHandlerInterface,
 } from '@internetarchive/shared-resize-observer';
+import { SortParam } from '@internetarchive/search-service/dist/src/search-params';
```
jbuckner, line 8: You should be able to import this just from `@internetarchive/search-service` without the full path. If you can't you may need to run `yarn` or rebase from `initial_work` to get the latest version of `search-service`

src/tiles/list/tile-list-compact.ts
```diff
@@ -1,20 +1,104 @@
...
     return html`
-      <h1>${this.model?.title}</h1>
-      <h2>${this.model?.datePublished?.toDateString()}</h2>
+      <div id="list-compact" class="${this.classSize}">
+        <div id="views">
+          ${formatCount(this.model?.viewCount ?? 0, this.formatSize)}
+        </div>
+        <div id="title">
+          <a
+            href="${this.baseNavigationUrl}/details/${this.model?.identifier}"
```
jbuckner, line 38: We are linking the entire cell inside an `<a>` tag from the `tile-dispatcher`... do you need to link this separately here?

src/tiles/list/tile-list-compact.ts
```diff
@@ -1,20 +1,104 @@
...
+      .desktop div {
+        font-size: 14px;
+      }
+
+      #list-compact {
+        display: grid;
+        grid-template-columns: 80px 3fr 115px 2fr 23.5px;
+        column-gap: 10px;
+
+        font-family: 'Helvetica Neue', Helvetica, Arial, sans-serif;
```
jbuckner, line 97: We have this font set at the top-level style. It should be getting inherited automatically. Is it not?

src/tiles/list/tile-list-compact.ts
```diff
@@ -1,20 +1,104 @@
...
+            title=${ifDefined(this.model?.title)}
+          >
+            ${this.model?.title}
+          </a>
+        </div>
+        <div id="date-published">${formatDate(this.date, this.formatSize)}</div>
+        <div id="creator">${this.model?.creator}</div>
+        <div id="icon">
+          <mediatype-icon
+            mediatype="${ifDefined(this.model?.mediatype)}"
```
jbuckner, line 48: When you're binding properties in Lit, you typically want to prefix the property name with a dot `.` so it uses Lit's binding which is much more powerful than html binding. ie:
```
```

src/mediatype-icon.ts
```diff
@@ -10,13 +10,17 @@ import { textsIcon } from './assets/img/icons/mediatype/texts';
 import { tvIcon } from './assets/img/icons/mediatype/tv';
 import { videoIcon } from './assets/img/icons/mediatype/video';
 import { webIcon } from './assets/img/icons/mediatype/web';
+import { collectionIcon } from './assets/img/icons/mediatype/collection';
 
 @customElement('mediatype-icon')
 export class MediatypeIcon extends LitElement {
   @property({ type: String }) mediatype = '';
 
+  @property({ type: String }) icontype = 'tile';
```
jbuckner, line 19: For things like this that have limited options, add a `type` to restrict the options, ie:
```
type IconType = 'list' | 'tile';
...
@property({ type: String }) icontype: IconType = 'tile';
```
This will give you type checking on it.

---

I also have another suggestion below about this property.

jbuckner (changes requested):

src/mediatype-icon.ts
```diff
@@ -44,12 +49,15 @@ export class MediatypeIcon extends LitElement {
   };
 
   render() {
-    return html`
-      <div>
-        ${this.mediatypeIcons[this.mediatype]}
-        <p class="status-text">${this.mediatypeText[this.mediatype]}</p>
-      </div>
-    `;
+    if (this.icontype === 'tile') {
```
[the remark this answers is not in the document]
mc2: Created `@property({ type: Boolean }) includeMediatypeText = true;`
to match `private readonly mediatypeText`

src/mediatype-icon.ts
mc2: having it restricted by the size of containing `<div>` seems to work. TBD.

src/tiles/tile-dispatcher.ts
mc2: Done. 

@jbuckner: note fuller path used in `app-root.ts` and `collection-browser.ts`

src/tiles/list/tile-list-compact.ts
mc2: Good question. Done that way in tile.inc. Design question. A user might right click on the explicit hyperlink as they aren't aware of the full cell link. We could just style this as a link, but if that, why not just keep it a link. Will ask in offshoot meeting.

src/tiles/list/tile-list-compact.ts
mc2: @jbuckner Let's talk about this. Getting an odd message. May need to change `mediatype-icon` slightly to handle `undefined`

src/mediatype-icon.ts
mc2: used boolean

src/tiles/list/tile-list-compact.ts
jbuckner: Ah interesting. Yeah I originally put it at the top so each of the tiles didn't have to do it, but if the list tiles aren't completely linked, we can remove it from the top-level and just wrap the grid tiles so the list tiles can do their own linking.

mc2: changing based on Jim Shelton in #iaux-collection-browser

src/tiles/list/tile-list-compact.ts
mc2: Changed mediatype-icon

src/models.ts
```diff
@@ -16,8 +16,11 @@ export interface TileModel {
   description?: string;
   collectionIdentifier?: string;
   collectionName?: string;
+  collections?: string[];
```
jbuckner, line 19: There are two `collections` arrays in here... also on line 21.

src/tiles/list/tile-list-compact.ts
```diff
@@ -1,22 +1,204 @@
...
+          ${formatCount(this.model?.viewCount ?? 0, this.formatSize)}
+        </div>
+        <div id="icon">
+          <mediatype-icon .mediatype=${this.model?.mediatype}></mediatype-icon>
+        </div>
+      </div>
     `;
   }
 
+  private img() {
```
jbuckner, line 39: Since this is basically a getter, you can use the getter syntax. Also, when you're returning a template like this one is, one good practice I've found is to add the suffix `Template`, ie:
```
private get imageTemplate() {
  return html`...`
}
```

src/tiles/list/tile-list-compact.ts
```diff
@@ -1,22 +1,204 @@
...
+    if (this.model?.identifier) {
+      return html` <img
+        src="${this.baseNavigationUrl}/services/img/${this.model.identifier}"
+        alt="${this.model.identifier}"
+        class=${this.model?.mediatype}
+      />`;
+    }
+    return html``;
```
jbuckner, line 47: If you're returning an empty template like this, it's better to use the `nothing` keyword so lit doesn't do any bindings. ie:
```
import { nothing } from 'lit';
...
return nothing;
```

src/tiles/list/tile-list-compact.ts
```diff
@@ -1,22 +1,204 @@
...
+      .desktop #icon {
+        height: 20px;
+      }
+
+      /* list-line */
+
+      #list-line {
+        display: grid;
+        column-gap: 10px;
+        border-top: 1px solid #ddd !important;
```
jbuckner, line 163: What is the `!#important` overriding?

src/tiles/list/tile-list-compact.ts
```diff
@@ -1,22 +1,204 @@
...
+        flex-shrink: 2;
+        flex-basis: 0;
+      }
+
+      #date {
+        flex-grow: 1;
+        flex-shrink: 1;
+        flex-basis: 0;
+      }
+      */
```
jbuckner, line 198: Do you need this commented style still?

src/tiles/list/tile-list.ts
```diff
@@ -0,0 +1,326 @@
...
+
+  render() {
+    return html`
+      <div id="list-line" class="${this.classSize}">
+        <div id="list-line-left">
+          <div id="thumb">${this.img()}</div>
+          ${this.iconLeft()}
+        </div>
+        <div id="list-line-right">
+          <div id="title">${DOMPurify.sanitize(this.model?.title ?? '')}</div>
```
dualcnhq, line 31: question:
should we add `dompurify` in all title and creator data?

src/tiles/list/tile-list.ts
```diff
@@ -0,0 +1,326 @@
...
+    }
+    return html``;
+  }
+
+  private searchLink(field: string, searchTerm: string) {
+    if (!field || !searchTerm) {
+      return html``;
+    }
+    const query = encodeURIComponent(`${field}:"${searchTerm}"`);
+    // eslint-disable-next-line lit/no-invalid-html
```
jbuckner, line 159: Q: what invalid html was this detecting?

src/tiles/list/tile-list.ts
```diff
@@ -0,0 +1,326 @@
...
+  private source() {
+    if (this.model?.source) {
+      return html`<div id="source">
+        <span class="label">Source: </span>${this.searchLink(
+          'source',
+          this.model.source
+        )}
+      </div>`;
+    }
+    return html``;
```
jbuckner, line 151: You can use `nothing` in this file where you're returning empty strings.

src/tiles/list/tile-list.ts
```diff
@@ -0,0 +1,326 @@
...
+    }
+    return html`
+      <div id="icon-left">
+        <mediatype-icon .mediatype=${this.model?.mediatype} .showText=${true}>
+        </mediatype-icon>
+      </div>
+    `;
+  }
+
+  private creator() {
```
jbuckner, line 71: Suggestion: use getter syntax for all of these templates and suffix the names with `Template`

src/tiles/list/tile-list.ts
mc2: Yes.

src/tiles/list/tile-list-compact.ts
```diff
@@ -1,22 +1,208 @@
...
+
+  @property({ type: Object }) sortParam?: SortParam;
+
   render() {
     return html`
-      <h1>${this.model?.title}</h1>
-      <h2>${this.model?.datePublished?.toDateString()}</h2>
+      <div id="list-line" class="${this.classSize}">
+        <div id="thumb">${this.imageTemplate}</div>
+        <div id="title">${DOMPurify.sanitize(this.model?.title ?? '')}</div>
```
dualcnhq, line 26: question: are we going to use dompurify for title and creator values now even on other tile components?

src/collection-browser.ts
```diff
@@ -1057,6 +1109,8 @@ export class CollectionBrowser
         --collectionBrowserCellMinHeight,
         5rem
       );
+      /* 30px, but compensating for a -5px margin */
+      --infiniteScrollerRowGap: 35px;
```
jbuckner, line 1113: Q: Where is that -5px margin coming from?

src/models.ts
```diff
@@ -7,21 +7,29 @@ export interface TileModel {
...
   viewCount: number;
+  volume?: string;
 }
 
-export type CollectionDisplayMode = 'grid' | 'list-compact' | 'list-detail';
+export type CollectionDisplayMode =
+  | 'grid'
+  | 'list-compact'
+  | 'list-detail'
+  | 'list-header';
```
jbuckner, line 32: Blocker: Can we decouple the tile type from the `CollectionDisplyMode`? This was intended to be used just for the 3 display modes we show the browser in. You could create a `TileType` that is basically this same thing and pass that into the dispatcher instead.

src/tiles/list/tile-list-compact-header.ts
```diff
@@ -0,0 +1,75 @@
...
+        <div id="date">Date Archived</div>
+        <div id="creator">Creator</div>
+        <div id="views">${eyeIcon}</div>
+        <div id="icon"></div>
+      </div>
+    `;
+  }
+
+  private get classSize(): string {
+    return (this.currentWidth ?? 531) < 530 ? 'mobile' : 'desktop';
```
jbuckner, line 29: Can you pass the [mobileBreakpoint](https://github.com/internetarchive/iaux-collection-browser/pull/10/files#diff-3f10ef47f877ef23eaab33c8a524ccd412dd0aa6ee7b067e94b28a2578943434R107) down into here instead of hardcoding the numbers?

src/tiles/list/tile-list-compact.ts
jbuckner: Might not be a bad idea to move it into the tile model level if we're going to have to sanitize it everywhere.

src/tiles/list/tile-list-compact.ts
```diff
@@ -1,22 +1,208 @@
...
+        <div id="date">${formatDate(this.date, this.formatSize)}</div>
+        <div id="creator">${DOMPurify.sanitize(this.model?.creator ?? '')}</div>
+        <div id="views">
+          ${formatCount(this.model?.viewCount ?? 0, this.formatSize)}
+        </div>
+        <div id="icon">
+          <mediatype-icon
+            .mediatype=${this.model?.mediatype}
+            style="--iconHeight: 20px; --iconWidth: 20px;"
```
jbuckner, line 35: Instead of inlining this, could you put this in the `static get styles()` section? ie:
```
mediatype-icon {
  --iconHeight: 20px; 
  --iconWidth: 20px;
}
```

src/tiles/list/tile-list-compact.ts
```diff
@@ -1,22 +1,208 @@
...
+          >
+          </mediatype-icon>
+        </div>
+      </div>
     `;
   }
 
+  private get imageTemplate() {
+    if (!this.model?.identifier) {
+      return nothing;
```
jbuckner, line 45: Should we have this default to that generic IA thumbnail image instead of removing the image altogether?

src/tiles/list/tile-list-compact.ts
```diff
@@ -1,22 +1,208 @@
...
+        return this.model?.dateReviewed;
+      case 'addeddate':
+        return this.model?.dateAdded;
+      default:
+        return this.model?.dateArchived; // publicdate
+    }
+  }
+
+  private get classSize(): string {
+    return (this.currentWidth ?? 531) < 530 ? 'mobile' : 'desktop';
```
jbuckner, line 73: Same thing about passing down the breakpoint here and just below.

src/tiles/list/tile-list-compact.ts
```diff
@@ -1,22 +1,208 @@
...
+        flex-shrink: 2;
+        flex-basis: 0;
+      }
+
+      #date {
+        flex-grow: 1;
+        flex-shrink: 1;
+        flex-basis: 0;
+      }
+    */
```
jbuckner, line 202: There's some commented css. Not sure if you still needed it

src/tiles/list/tile-list.ts
```diff
@@ -0,0 +1,374 @@
...
+  private get iconLeftTemplate() {
+    if (this.classSize !== 'desktop') {
+      return nothing;
+    }
+    return html`
+      <div id="icon-left">
+        <mediatype-icon
+          .mediatype=${this.model?.mediatype}
+          .showText=${true}
+          style="--iconHeight: 20px; --iconWidth: 20px;text-align: center;"
```
jbuckner, line 73: Could you move these inlines to the `styles()` section?

src/tiles/list/tile-list.ts
```diff
@@ -0,0 +1,374 @@
...
+        return this.model?.dateReviewed;
+      case 'addeddate':
+        return this.model?.dateAdded;
+      default:
+        return this.model?.dateArchived; // publicdate
+    }
+  }
+
+  private get classSize(): string {
+    return (this.currentWidth ?? 531) < 530 ? 'mobile' : 'desktop';
```
jbuckner, line 242: Pass in the mobile breakpoint here.

src/tiles/list/tile-list.ts
```diff
@@ -0,0 +1,374 @@
...
+  }
+
+  private get formatSize(): DateFormat | NumberFormat {
+    return (this.currentWidth ?? 531) < 530 ? 'short' : 'long';
+  }
+
+  static get styles() {
+    return css`
+      html {
+        font-size: unset;
```
jbuckner, line 252: Does this affect the font size at the page level or just inside this component?

jbuckner (changes requested):
Looking really good! A few things I noticed in here, but it's getting really close!

src/tiles/list/tile-list.ts
mc2: just inside this component

src/tiles/list/tile-list.ts
mc2: done

src/tiles/list/tile-list.ts
mc2: done

src/tiles/list/tile-list-compact.ts
mc2: done

src/collection-browser.ts
mc2: Added comment

src/tiles/list/tile-list.ts
mc2: Turns out is was the "" in query inside the "" for href=""
Fixed

src/tiles/list/tile-list-compact.ts
mc2: removed

src/tiles/list/tile-list-compact.ts
mc2: removed

src/tiles/list/tile-list-compact.ts
mc2: thanks, done

src/tiles/list/tile-list-compact.ts
mc2: done

src/models.ts
mc2: fixed

src/tiles/list/tile-list-compact.ts
jbuckner: In this case, the `TileModel` is already the view layer since it's purpose is specifically to back the UI tiles with "raw data" from search-service `Metadata`.

jbuckner: Actually, thinking more I think this is fine to do it at view time 👍

jbuckner (approved):
Nice! LGTM :shipit:

PR #11, opened 2022-03-21
fix mediatype-icons to showtext and allow undefined key

mc2:


src/mediatype-icon.ts
```diff
@@ -13,7 +13,11 @@ import { webIcon } from './assets/img/icons/mediatype/web';
 
 @customElement('mediatype-icon')
 export class MediatypeIcon extends LitElement {
-  @property({ type: String }) mediatype = '';
+  @property({ type: String }) mediatype: string | undefined;
+
+  @property({ type: Boolean }) showText = false;
+
+  // @property({ type: Boolean }) isColored = false;
```
jbuckner, line 20: You can remove this since it's not being used right now

jbuckner (approved):
One thing, but looks good 👍

src/mediatype-icon.ts
mc2: removed

PR #16, opened 2022-04-12
Add fill color mediatype icon

dualcnhq:
## Usage:

```
<mediatype-icon 
  ?showText=${true}
  style="--iconFillColor: #0066cc;"
>
</mediatype-icon>
```

## Sample:
Screen Shot 2022-04-12 at 17 30 48


jbuckner (approved):
That works! 👍

mc2:
I don't know that we need this -- all the spec'ed colors are black.
I do need a way to set the height of the icon, when the showText is true, i.e. separately from the text.

PR #17, opened 2022-04-12
Add width height mediatype icon

dualcnhq:
## Usage:

```
<mediatype-icon 
   ?showText=${true}
   style="--iconHeight: 10px; --iconWidth: 10px;"
>
</mediatype-icon>
```

## Sample:
Screen Shot 2022-04-13 at 01 13 23

### DIFF
--- a/src/mediatype-icon.ts
+++ b/src/mediatype-icon.ts
@@ -74,6 +74,11 @@ export class MediatypeIcon extends LitElement {
         display: none;
       }
 
+      svg {
+        height: var(--iconHeight);
+        width: var(--iconWidth);
+      }
+
       .fill-color {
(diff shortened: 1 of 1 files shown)


src/mediatype-icon.ts
```diff
@@ -74,6 +74,11 @@ export class MediatypeIcon extends LitElement {
         display: none;
       }
 
+      svg {
+        height: var(--iconHeight);
+        width: var(--iconWidth);
```
jbuckner, line 79: What happens when there are no values set for this?

dualcnhq: it uses the default size it is defined from the svg or the component

jbuckner (approved):
Looks good!

PR #18, opened 2022-04-12
Item tile improvements

dualcnhq:
## Changes:

- Add mediatype-icon height and width style
- Add default mediatype-icon height and width style 
- Add mediatype-icon fillColor
- Initial changes for item-tile styling


src/tiles/grid/item-tile.ts
```diff
@@ -71,6 +71,7 @@ export class ItemTile extends LitElement {
               <mediatype-icon
                 .mediatype=${this.model?.mediatype}
                 ?showText=${true}
+                style="--iconHeight: 10px; --iconWidth: 10px;"
```
jbuckner, line 74: Can we put this in the CSS `style` section instead of inline?

src/mediatype-icon.ts
```diff
@@ -41,21 +46,44 @@ export class MediatypeIcon extends LitElement {
...
     movies: 'Movie',
+    radio: 'Radio',
     software: 'Software',
     texts: 'Text',
     tv: 'TV',
     video: 'Video',
     web: 'Web',
   };
 
+  private getDisplayMediaType = () => {
```
jbuckner, line 60: For things that don't require parameters, I would suggest using the `getter` form:
```
private get displayMediaType() {
  ...
  return ...
}

...

const displayType = this.displayMediaType
```

src/tiles/grid/item-tile.ts
```diff
@@ -331,7 +340,8 @@ export class ItemTile extends LitElement {
         color: #333;
         text-decoration: none;
         display: block;
-      }
+      } 
+      */
```
jbuckner, line 344: Do you still need this commented style?

src/mediatype-icon.ts
```diff
@@ -3,31 +3,36 @@ import { customElement, property } from 'lit/decorators.js';
...
 import { tvIcon } from './assets/img/icons/mediatype/tv';
 import { videoIcon } from './assets/img/icons/mediatype/video';
 import { webIcon } from './assets/img/icons/mediatype/web';
-import { collectionIcon } from './assets/img/icons/mediatype/collection';
 
 @customElement('mediatype-icon')
 export class MediatypeIcon extends LitElement {
   @property({ type: String }) mediatype: string | undefined;
 
+  @property({ type: Array }) collection: string[] | undefined;
```
jbuckner, line 22: I might pluralize this since it's an array

src/mediatype-icon.ts
```diff
@@ -44,21 +49,44 @@ export class MediatypeIcon extends LitElement {
...
+    }
+    return this.mediatype;
+  };
+
   render() {
     if (!this.mediatype) {
       return html``;
     }
+
+    const displayMediatype = this.getDisplayMediaType() || '';
```
jbuckner, line 84: Does `getDisplayMediaType()` return an optional?

src/mediatype-icon.ts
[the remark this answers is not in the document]
dualcnhq: as of now, i put it here since it only affects the mediatype icon display

src/mediatype-icon.ts
dualcnhq: i'm getting:
```
const displayMediatype: string | undefined
Type 'undefined' cannot be used as an index type.ts(2538)
```

if i don't add `|| ''` in the end

src/tiles/grid/item-tile.ts
```diff
@@ -76,12 +73,17 @@ export class ItemTile extends LitElement {
             <span>${itemCreator}</span>
           </div>
         </div>
+
+        <div class="hr"></div>
```
jbuckner, line 77: I think my other comment got lost in the rebase, but do you need a separate div for this horizontal line or can you set a `border-top` on the `#item-stats-container`?

jbuckner (approved):
Looks good 👍

PR #21, opened 2022-04-26
Date sort display, short date format

mc2:
DRY date-label method
Add display of date label + value if sorted on that date type in detailed list tile
Change short date format from 'yyyy' to 'Jan yyyy'


jbuckner (approved):

dualcnhq (approved):
LGTM

PR #23, opened 2022-04-26
Fix item creator text alignment to start from bottom

dualcnhq:


src/tiles/grid/item-tile.ts
```diff
@@ -223,19 +218,21 @@ export class ItemTile extends LitElement {
...
         -webkit-box-orient: vertical;
+        display: flex;
+        justify-content: center;
+        align-items: flex-end;
+        padding-top: 1rem;
       }
 
-      #text-by {
+      .item-creator span:before {
+        content: 'By: ';
```
jbuckner, line 235: This feels wrong to put semantically meaningful content in your CSS. What effect are you trying to achieve with this?

dualcnhq: i tried to separate the span text but got different positions on each so i did it this way.

is it better if i change it to: `<span><strong>By:</strong>${itemCreator}</span>`

jbuckner: Yes, that would be better

dualcnhq: updated...

src/tiles/grid/item-tile.ts
```diff
@@ -223,19 +218,20 @@ export class ItemTile extends LitElement {
...
         -webkit-line-clamp: 2;
         -webkit-box-orient: vertical;
+        display: flex;
+        justify-content: center;
+        align-items: flex-end;
+        padding-top: 1rem;
       }
 
-      #text-by {
+      #by {
```
jbuckner, line 234: Does this `#by` element exist? I didn't notice it above.

dualcnhq: good catch... i just removed it

jbuckner (approved):
Looks good! 👍

PR #25, opened 2022-04-27
Account list tiles

mc2:
Add "Archivist since" label in creator slot


mc2:
test `query=collection%3Afav-markc2` look for "Katie Barrett"

jbuckner (approved):
Looks good 👍

PR #26, opened 2022-04-28
Move mediatype-colors and text to separate file

dualcnhq:
Changes:

- fix item creator text with ellipsis
- combine mediatype data to 1 object import
- add `ifDefined` directive to load mediatype data


src/mediatype/mediatype-color.ts
```diff
@@ -0,0 +1,14 @@
+export const mediatypeColor: { [key: string]: any } = {
```
mc2, line 1: Suggestion: Add:
collection, radio, tv

src/mediatype-icon.ts
```diff
@@ -82,11 +68,16 @@ export class MediatypeIcon extends LitElement {
     }
 
     const displayMediatype = this.displayMediaType || '';
+    const fillColor = mediatypeColor[displayMediatype];
```
mc2, line 71: Question: What happens when `mediatypeColor[displayMediatype]` doesn't exist? When `displayMediatype` is ""?
Does the style attribute get removed?

src/mediatype-icon.ts
```diff
@@ -82,11 +68,16 @@ export class MediatypeIcon extends LitElement {
...
 
     return html`
-      <div id="icon" class="${this.showText ? 'show-text' : 'hide-text'}">
+      <div
+        id="icon"
+        class="${this.showText ? 'show-text' : 'hide-text'}"
+        style="--iconFillColor:${fillColor}"
```
mc2, line 77: Suggestion: Use https://lit.dev/docs/templates/directives/#ifdefined, ifDefined directive to: 
- allow fillcolor to be undefined (I am not clear about the state of the `fillColor` when `displayMediatype` isn't a key match in `mediatypeColor`)
- remove style attribute

src/mediatype/mediatype-text.ts
```diff
@@ -0,0 +1,16 @@
+export const mediatypeText: { [key: string]: any } = {
```
mc2, line 1: Suggestion: A ... design, software maintenance goal would be to have all the mediatype-related information in one place. So that adding a mediatype or changing a mediatype has one single place: source of truth.

You might have a single object with three properties, icon, text, color

src/mediatype-icon.ts
dualcnhq: no, it stays there with no value.
same with the icons and icon text

src/mediatype/mediatype-display.ts
```diff
@@ -0,0 +1,31 @@
+export const mediatypeColor: { [key: string]: string } = {
```
mc2, line 1: suggestion: 

```
export const mediatypeData: { [key: string]: string } = {
   account: {
     color: '#000000',
     icon: accountIcon,
     text: 'Account',
   }, 
   audio: { ... }. ...
   ```

Comment (dualcnhq) [reply src/mediatype/mediatype-display.ts]: PR updated

File: src/tiles/grid/item-tile.ts [hunk 6]
```diff
@@ -217,19 +202,33 @@ export class ItemTile extends LitElement {
...
+        text-overflow: ellipsis;
+        overflow: hidden;
+        display: -webkit-box;
+        -webkit-box-orient: vertical;
+        word-wrap: break-word;
+        line-height: 2rem;
+        text-align: center;
+      }
+
+      .hr {
```
dualcnhq, line 230: added this for now to test in offshoot how it looks like

src/mediatype-icon.ts
```diff
@@ -81,12 +67,20 @@ export class MediatypeIcon extends LitElement {
...
+    const displayMediatype = this.displayMediaType ?? '';
 
     return html`
-      <div id="icon" class="${this.showText ? 'show-text' : 'hide-text'}">
-        ${this.mediatypeIcons[displayMediatype]}
-        <p class="status-text">${this.mediatypeText[displayMediatype]}</p>
+      <div
+        id="icon"
+        class="${this.showText ? 'show-text' : 'hide-text'}"
+        style="--iconFillColor: ${ifDefined(
```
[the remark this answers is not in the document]
mc2: Oops: I forget I have to click the review button when pending.

I would try this and, ideally, add tests. But you could remove line 39 probably. Again try a test. :)

src/mediatype-icon.ts
```diff
@@ -1,19 +1,8 @@
...
-import { etreeIcon } from './assets/img/icons/mediatype/etree';
-import { imagesIcon } from './assets/img/icons/mediatype/images';
-import { filmIcon } from './assets/img/icons/mediatype/film';
-import { radioIcon } from './assets/img/icons/mediatype/radio';
-import { softwareIcon } from './assets/img/icons/mediatype/software';
-import { textsIcon } from './assets/img/icons/mediatype/texts';
-import { tvIcon } from './assets/img/icons/mediatype/tv';
-import { videoIcon } from './assets/img/icons/mediatype/video';
-import { webIcon } from './assets/img/icons/mediatype/web';
+import { mediatypeData } from './mediatype/mediatype-display';
```
mc2, line 5: Blocker: By Lit convention, the class name should match the file name.
So, asked around about a good name, suggestion: 

`import { mediatypeConfig } from './mediatype/mediatype-config';`

src/mediatype-icon.ts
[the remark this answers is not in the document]
mc2: https://lit.dev/docs/templates/directives/#ifdefined -- 
handles undefined and null -- but I really need to test when 

`mediatypeConfig[undefined].Color`
is the result of 
`mediatypeConfig[this.displayMediaType].Color`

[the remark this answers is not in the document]
dualcnhq: `mediatypeConfig` only accepts `string` type as parameter and complains if parameter is `undefined`

src/mediatype-icon.ts
mc2: Please rename either file or name of class to match Lit convention: 
camel case class name => lowercase, -[dash] separated file name.
 so 
 mediatypeData => mediatype-data
 But suggest name change of both: 
 mediatypeConfig ...

mc2 (changes requested):

src/mediatype-icon.ts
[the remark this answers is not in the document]
jbuckner: You can do `if (!this.mediatype || !this.displayMediaType) return;` above so you only show it if it's defined.

src/tiles/grid/item-tile.ts
```diff
@@ -85,18 +67,21 @@ export class ItemTile extends LitElement {
...
           <div id="item-image-container">${this.renderItemImageView}</div>
           <div class="item-creator">
-            <span><strong>By:&nbsp;</strong>${itemCreator}</span>
+            <div class="truncated">
+              <span><strong>By:&nbsp;</strong>${itemCreator}</span>
+            </div>
           </div>
         </div>
 
+        <div class="hr"></div>
```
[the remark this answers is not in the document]
dualcnhq: yup still being used. i added it for now to check how it looks like in offshoot

dualcnhq:
> You can do if (!this.mediatype || !this.displayMediaType) return; above so you only show it if it's defined.

uhmmm... i think the point of having `ifDefined` directive is to check if the value from `mediatypeConfig[key]` (wherein key is the `mediatype`) is `defined` or `undefined`

if it's `undefined` then there will be nothing to render and the attributes will be blank, but if it's defined or exists then load it.

not sure of the proper way/syntax to do it, but it does the job  ¯\_(ツ)_/¯

jbuckner:
> uhmmm... i think the point of having ifDefined directive is to check if the value from mediatypeConfig[key] (wherein key is the mediatype) is defined or undefined
> if it's undefined then there will be nothing to render and the attributes will be blank, but if it's defined or exists then load it.

Yes `ifDefined` will not include it if it's `undefined`, but in this case, the error you are seeing is coming before it ever gets to `ifDefined()`. The issue is that `this.displayMediaType` is potentially `undefined` and you cannot do a lookup in a JSON object for an `undefined` object. ie this isn't valid: `myObject[undefined]`.

While your solution does solve the issue, it does so by using a side effect of the lookup since `myObject['']` will never return anything.

Additionally, the [current early return](https://github.com/internetarchive/iaux-collection-browser/pull/26/files#diff-a5967a0f3f427e4856ece37e2f7b4748a2a1df8ae5a51cbcd96fb8247d7283d6R35) that you have `if (!this.mediatype) { return html`` }` is no longer being used because you're not referencing it anywhere in this method. If you update that check to `if (!this.displayMediaType)`, which is what you're actually using, it solves the problem.

mc2:
Oops! Did the merge

PR #30, opened 2022-05-01
Mobile list, collections as promise[]

mc2:
Complete mobile view of detailed list view.
Convert collections links changing async component to async state and function

### DIFF
--- a/src/tiles/list/tile-list.ts
+++ b/src/tiles/list/tile-list.ts
@@ -1,8 +1,15 @@
-import { css, html, LitElement, nothing } from 'lit';
+import {
+  css,
+  html,
+  LitElement,
+  nothing,
+  PropertyValues,
+  TemplateResult,
+} from 'lit';
 import { ifDefined } from 'lit/directives/if-defined.js';
 import { join } from 'lit/directives/join.js';
 import { map } from 'lit/directives/map.js';
-import { customElement, property } from 'lit/decorators.js';
+import { customElement, property, state } from 'lit/decorators.js';
 import { SortParam } from '@internetarchive/search-service';
 import DOMPurify from 'dompurify';
 import { CollectionNameCacheInterface } from '@internetarchive/collection-name-cache';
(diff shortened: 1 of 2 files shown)


src/tiles/list/tile-list.ts
```diff
@@ -30,9 +37,54 @@ export class TileList extends LitElement {
...
+    }
+  }
+
+  private async fetchCollectionNames() {
+    if (
+      !this.model?.collections ||
+      this.model.collections.length === 0 ||
+      !this.collectionNameCache
+    ) {
+      return [];
```
jbuckner, line 54: This method doesn't have a return type.. do you need to return these or can you just return? Same down on line 72.

src/tiles/list/tile-list.ts
```diff
@@ -410,11 +455,18 @@ export class TileList extends LitElement {
         line-height: 20px;
       }
 
+      #description,
       #creator,
-      #topic,
+      #topics,
       #source {
```
jbuckner, line 461: Q: I noticed you added a `metadata` class to some of these elements. Did you mean for this to be `.metadata {` since it looks like you're applying this to the metadata elements?

If not, may I suggest using a class since you're applying this style to several elements?

jbuckner (commented):
Two questions, but looking good!

src/tiles/list/tile-list.ts
mc2: A: The metadataTemplate() has and `id?` optional parm just so it can add an id= attribute to a few select elements. I removed the id attribute from all other "metadata-display" type fields.

For these given an id, I anticipate that the truncate/ellipsis will be adjusted, perhaps differently in mobile view.
I considered creating different "truncate-3" and truncate-2" (-line) classes, but I retained the id= method for flexibility and precise identification, i.e. seeing in the CSS that these and only these have this truncated behavior. Thus, not having to look in the template creation code to set/change the class/display behavior.

If you think this unwise, please let me know.

mc2: "source" seems that it could be truncated to 1 line, for example. Meeting with Jim S. soon to model/discuss.

mc2: @jbuckner I'll let you "Resolve" this.

jbuckner: This was more of just a general comment about applying the same style to many elements. That's typically what classes are used for, but feel free to do it however you see fit.

jbuckner (approved):

PR #31, opened 2022-05-02
Update mediatype icon

dualcnhq:


src/mediatype-icon.ts
```diff
@@ -32,7 +32,7 @@ export class MediatypeIcon extends LitElement {
   }
 
   render() {
-    if (!this.mediatype) {
+    if (!this.mediatype || !this.displayMediatype) {
```
jbuckner, line 35: You don't need the `!this.mediatype` check anymore since you're not referencing it in the `render()` method.

jbuckner:
Could you also DRY up the render method a bit? I was going to comment on the other ticket... you call `mediatypeConfig[this.displayMediatype]` several times... you can just check for that instead of doing all of those `ifDefined()`:

```
${ifDefined(
          mediatypeConfig[this.displayMediatype].color
        )}"
      >
        ${ifDefined(mediatypeConfig[this.displayMediatype].icon)}
        <p class="status-text">
          ${ifDefined(mediatypeConfig[this.displayMediatype].text)}
        </p>
```

src/mediatype-icon.ts
```diff
@@ -32,22 +32,20 @@ export class MediatypeIcon extends LitElement {
   }
 
   render() {
-    if (!this.mediatype) {
+    if (!this.displayMediatype) {
       return html``;
     }
 
+    const toDisplayMediatype = mediatypeConfig[this.displayMediatype];
```
jbuckner, line 39: Suggestion.. since this is the config, I would name it as such. The display media type is the thing you're using to look up the config.

Also, If you do this just below, you can get rid of those `ifDefined()` in the HTML since it will always be defined:
```
if (!toDisplayMediaType) { return html`` }
```

dualcnhq: updated 🙇

jbuckner (approved):

PR #33, opened 2022-05-02
Set list-only collection icon color

mc2:
Added --iconCustomFillColor

### DIFF
--- a/src/tiles/list/tile-list-compact.ts
+++ b/src/tiles/list/tile-list-compact.ts
@@ -1,4 +1,5 @@
 import { css, html, LitElement, nothing } from 'lit';
+import { ifDefined } from 'lit/directives/if-defined.js';
 import { customElement, property } from 'lit/decorators.js';
 import { SortParam } from '@internetarchive/search-service';
 import DOMPurify from 'dompurify';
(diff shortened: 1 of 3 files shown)


src/tiles/list/tile-list-compact.ts
```diff
@@ -43,6 +49,14 @@ export class TileListCompact extends LitElement {
     `;
   }
 
+  // Only in list, not tile
+  private get collectionColor() {
```
dualcnhq, line 53: q: i see that `collectionColor` is defined here and also in `src/tiles/list/tile-list.ts`, can it be just 1?

mc2: A: great question! In order for it to be a `get`, I think it would have to be in a 'list' superclass and inherited. I could make it a directive, or an outside function like `src/tiles/list/account-label.ts`

I agree it should be DRY'ed up. I'll take a look at a directive again, to see it it's appropriate

jbuckner: Is there a reason we're using a different color for the collection icon on these tiles?

mc2: There are a number of repeated elements between 'compact' and 'detailed' lists where it seems a mixin would work to remove the duplications -- i.e. a 'list' superclass, in essence, as noted above

mc2: https://lit.dev/docs/composition/mixins/#mixins-in-typescript

jbuckner (approved):

PR #34, opened 2022-05-03
Refactoring + waveform thumbnail changes

dualcnhq:
Changes:

- Create item-image component for default and deemphasize tile image load
- Waveform thumbnail refactor


src/tiles/item-image.ts
```diff
@@ -0,0 +1,206 @@
...
+  }
+
+  private setDeemphsize() {
+    this.isDeemphasize =
+      this.model?.collections.includes('deemphasize') ?? false;
+  }
+
+  render() {
+    return html`
+      <div class=${ifDefined(this.imageBoxClass)}>
```
dualcnhq, line 37: i added `ifDefined` so in the DOM it will not be: `<div class>`

mc2 (approved):
Looks good to me

src/tiles/item-image.ts
mc2: Great, thanks!

PR #35, opened 2022-05-04
Item tile update

dualcnhq:


src/tiles/item-image.ts
```diff
@@ -21,6 +21,8 @@ export class ItemImage extends LitElement {
 
   @state() private isDeemphasize = false;
 
+  @query('.item-image') _itemImageWaveform!: HTMLImageElement;
```
jbuckner, line 24: Suggestion: use `private` instead of prefixing with an underscore. You don't really need to use the underscore private convention since typescript natively supports `private` properties.

jbuckner (approved):
One suggestion, but looks good otherwise

PR #39, opened 2022-05-05
Use item-image in lists

mc2:
[WEBDEV-5248](https://webarchive.jira.com/browse/WEBDEV-5248)
add deemphasize blur to list tiles

Created item-image component.
Changed list tiles to use that component.
Kept code design so that outside container controls image size, border-radius.
Might prefer deemphasize logic be part of model, logic outside scope of image display
I think this component has gotten too complicated

Not using `background-image:url()` in order to use alt="" Don't see why background method used.


src/tiles/item-image.ts
```diff
@@ -50,12 +52,32 @@ export class ItemImage extends LitElement {
   }
 
   // Templates
-  private get backgroundImageTemplate() {
+  private get ImageTemplate() {
```
bfalling, line 55: QUESTION: Is there a reason this starts with a capital letter? Seems like it should be lowercase.

src/tiles/list/tile-list-compact.ts
```diff
@@ -63,11 +65,14 @@ export class TileListCompact extends LitElement {
...
-      src="${this.baseImageUrl}/services/img/${this.model.identifier}"
-      alt="${this.model.identifier}"
-      class="${this.model?.mediatype}"
-    />`;
+    return html`
+      <item-image
+        .model=${this.model}
+        .baseImageUrl=${this.baseImageUrl}
+        .isListTile=${true}
+      >
```
bfalling, line 73: QUESTION: Is it desired to have a unit of whitespace between the opening and closing `item-image` tags?

bfalling (commented):
A few questions. Otherwise LGTM.

src/tiles/item-image.ts
mc2: fixed

src/tiles/list/tile-list-compact.ts
mc2: The indentation and end tag on a separate line following examples in lit.dev and @jbuckner's code.
I like it for readability.
I have found whitespace problems in HTML display only when a component wraps another component.

src/tiles/item-image.ts
```diff
@@ -52,12 +56,29 @@ export class ItemImage extends LitElement {
...
+      ${this.tileActionTemplate}
+    `;
+  }
+
+  private get listImageTemplate() {
+    if (!this.model) {
+      return nothing;
+    }
+    return html`
+      <img src="${this.imageSrc}" alt="" class="${this.listImageClass}" />
```
dualcnhq, line 80: is alt still blank?

dualcnhq: other than that, LGTM

mc2: Yes, we discussed this - not having a screen reader repeat the title for a user.

mc2: We might do this for the other images, I.e., waveforms, line 91

dualcnhq: is it possible to have a generic name as `alt` in case?

iisa: ...hierarchy-wise, it would be nice to have the alt here instead of the image. bc when u source svg using img tag, the title may not come thru as opposed to having it on the img tag straight up.

src/tiles/item-image.ts
```diff
@@ -19,30 +19,34 @@ export class ItemImage extends LitElement {
...
+      this.setDeemphasize();
     }
   }
 
-  private setDeemphsize() {
+  // Don't deemphasize if item is a collection
+  private setDeemphasize() {
     this.isDeemphasize =
-      this.model?.collections.includes('deemphasize') ?? false;
+      (this.model?.mediatype !== 'collection' &&
```
iisa, line 39: Q: do you need this part of the check `this.model?.mediatype !== 'collection'`?

mc2: Yes

PR #42, opened 2022-05-05
Fix audio mediatype background with waveform image

dualcnhq:


src/tiles/item-image.ts
```diff
@@ -86,6 +88,7 @@ export class ItemImage extends LitElement {
       this.itemImageWaveform.naturalWidth /
       this.itemImageWaveform.naturalHeight;
     if (aspectRatio === 4) {
+      this.box.classList.add(this.randomGradient);
```
jbuckner, line 91: Instead of manipulating classes directly, I would recommend setting a state property and change your classes based on the state. ie:
```ts
@state() isWaveform: boolean = false;
...
if (aspectRatio === 4) this.isWaveform = true;
...
<div class="box ${this.isWaveform ? this.randomGradient : ''}">
```
Similar comment for the line just below this line.

src/tiles/item-image.ts
```diff
@@ -61,7 +63,7 @@ export class ItemImage extends LitElement {
 
   private get waveformTemplate() {
     return html`
-      <div class="box" id=${this.randomGradient}>
+      <div class="box">
```
jbuckner, line 66: Since there's only one of these elements and you're specifically looking for it above, I would make this an `id` instead of a class. Classes are for when there are many elements of a similar style.

Also, `box` is a very generic name.. could you use something more specific?

jbuckner (approved):

PR #44, opened 2022-05-05
Add default value for baseImageUrl

dualcnhq:


src/collection-browser.ts
```diff
@@ -1048,7 +1048,7 @@ export class CollectionBrowser
 
     return html` <tile-dispatcher
       .baseNavigationUrl=${this.baseNavigationUrl}
-      .baseImageUrl=${this.baseImageUrl}
+      .baseImageUrl=${this.baseImageUrl ?? 'https://archive.org'}
```
jbuckner, line 1051: Set this where the actual property is defined instead of where it's being passed in

jbuckner (approved):

PR #48, opened 2022-05-06
Update generating gradient background for waveform thumbnails

dualcnhq:


src/tiles/item-image.ts
```diff
@@ -110,7 +110,13 @@ export class ItemImage extends LitElement {
   }
 
   private get randomGradient() {
-    return `grad${Math.floor(Math.random() * (6 - 1) + 1)}`;
+    return `grad${this.generateHashCode(this.model?.identifier!) % 6}`;
```
jbuckner, line 113: Blocker: Don't force unwrap an optional value (don't use the exclamation mark). This can cause exceptions in runtime code.

Also, could you split this line up into one that generates your number.. there's a lot going on in this one line.

jbuckner (approved):

PR #52, opened 2022-05-11
Item image updates

dualcnhq:
## Changes:

- Add etree mediatype checking for waveform thumbnail background
- Change waveform styling and deemphasize background
- Decouple item-image component

TODO:
- implement changes for item-image in tile-list component as well
- add dynamic sizing for the item-image


src/tiles/item-image.ts
```diff
@@ -36,10 +39,18 @@ export class ItemImage extends LitElement {
...
+      this.isWithWaveformMediatype = withWaveformMediatype.includes(
+        this.model?.mediatype
+      );
+  }
+
   render() {
     return html`
       <div class=${ifDefined(this.imageBoxClass)}>
-        ${this.model?.mediatype === 'audio'
+        ${this.isWithWaveformMediatype
```
iisa, line 53: you can do a getter here do you don't need to bloat state, like: 
```
get isWithWaveformMediatType() {
return this.model?.mediatype === 'audio'
}
```

src/tiles/item-image.ts
```diff
@@ -38,201 +29,41 @@ export class ItemImage extends LitElement {
...
-  private get backgroundImageTemplate() {
-    return html`
-      <div
-        class=${this.imageClass}
-        style="background-image:url(${this.imageSrc})"
-      ></div>
-    `;
+  private get isWithWaveformMediatype() {
+    return (
+      this.model?.mediatype === 'audio' || this.model?.mediatype === 'etree'
```
iisa, line 44: wow, TIL `etree` is a mediatype

iisa (commented):
thank you for the care in separating image logic. 👍 

will look for unit tests & qa steps

src/tiles/item-image.ts
iisa: Q: some `etree` / lam items have primary images. what shows up on their tile? waveform bg w/ primary image?

dualcnhq:
closing this branch - moving this to: https://github.com/internetarchive/iaux-collection-browser/pull/57

src/tiles/item-image.ts
dualcnhq: were currently just detecting if the loaded image is a waveform based on the aspectRation... so if we detected it's a waveform, then we'll load the waveform with background, but if it has an actual image cover/album cover of some sort then we'll load that image instead

PR #53, opened 2022-05-16
Add gh-pages + vite + npm build

dualcnhq:
### Changes:

- add `gh-pages` for deployment and review app QA testing
- add [vitejs](https://vitejs.dev/) for creating bundled app

### Usage (Manual step):

- install deps: `npm install`
- deploy to GH pages:  `npm run deploy -- -e <branch_name>`
- add/configure HMR

TODO:
- add github workflow to automate build process and deployment per branch


.gitignore
```diff
@@ -1,6 +1,13 @@
-## editors
-/.idea
-/.vscode
+# Editor directories and files
+.vscode/*
+!.vscode/extensions.json
+.idea
+.DS_Store
```
mc2, line 5: Remove duplicate: See below line 13, in correct place under `## system files`

tsconfig.json
```diff
@@ -10,11 +10,12 @@
     "allowSyntheticDefaultImports": true,
     "experimentalDecorators": true,
     "importHelpers": true,
-    "outDir": "dist",
+    "outDir": "build",
```
mc2, line 13: Question: why the directory change?

README.md
```diff
@@ -13,53 +13,60 @@ import '@internetarchive/collection-browser';
 </collection-browser>
 ```
 
-## Local Demo with `web-dev-server`
+## Local Demo with `vite`
 ```bash
-yarn start
```
mc2, line 18: Question: Why are we changing from yarn to npm?

mc2 (changes requested):

README.md
dualcnhq: i figured if we can use npm vs yarn now on our projects, but can change it back to yarn

tsconfig.json
dualcnhq: since we are now using vite for bundling to prod-ish and development, i figured we can use the same directory for it, but we could change it to be both `dist` instead if we want to

README.md
mc2: @dualcnhq Re. "why?" 
I was looking for an answer like, "We're trying to standardize all our JS projects on ..."
I remember the question of 'npm vs yarn" being asked in a recent meeting -- but I don't remember a firm decision being made. 
Do at least @iisa and @traceypooh agree on a direction?

tsconfig.json
mc2: I am not familiar with the interaction of build tools enough to know what to check for. 

For example, in `tsconfig.json` we have: 
```
{
  "compilerOptions": {
     "outDir": "dist",
     ...
```
Are we going to have both directories?

mc2: I also see a failing build step at: 
```
Error: Could not find any test files with pattern(s): dist/test/**/*.test.js
```

dualcnhq: seems we need both directories, `build` for gh-pages publishing; `dist` for tests

README.md
iisa: let's move back to yarn to stay steady until release.  Vite looks compatible w yarn

dualcnhq:
closing this to use rollup instead - https://github.com/internetarchive/iaux-collection-browser/pull/58

tsconfig.json
dualcnhq: yup, moved now

PR #54, opened 2022-05-18
Add logged in and content warning to model

mc2:
Our previous understanding of "deemphasize" has changed.

The collection `loggedin` flags tiles to be blurred and a different warning for items depending if user logged in or not.
The collection `no-preview` flags a blur and content warning.

Moved determination of these states to model and added properties to model.

Created "Simulate Login" to allow testing in demo.

### DIFF
--- a/demo/app-root.ts
+++ b/demo/app-root.ts
@@ -32,7 +32,7 @@ export class AppRoot extends LitElement {
 
   @state() private colGap: number = 1.7;
 
-  @state() private LoggedIn: boolean = false;
+  @state() private loggedIn: boolean = false;
 
   @query('#base-query-field') private baseQueryField!: HTMLInputElement;
 
@@ -170,7 +170,7 @@ export class AppRoot extends LitElement {
           .resizeObserver=${this.resizeObserver}
           .collectionNameCache=${this.collectionNameCache}
           .showHistogramDatePicker=${true}
-          .loggedIn=${this.LoggedIn}
+          .loggedIn=${this.loggedIn}
           @visiblePageChanged=${this.visiblePageChanged}
           @baseQueryChanged=${this.baseQueryChanged}
         >
@@ -186,9 +186,9 @@ export class AppRoot extends LitElement {
   private loginChanged(e: Event) {
     const target = e.target as HTMLInputElement;
     if (target.checked) {
-      this.LoggedIn = true;
+      this.loggedIn = true;
     } else {
-      this.LoggedIn = false;
+      this.loggedIn = false;
     }
   }
 
--- a/demo/app-root.ts
+++ b/demo/app-root.ts
@@ -32,6 +32,8 @@ export class AppRoot extends LitElement {
 
   @state() private colGap: number = 1.7;
 
+  @state() private LoggedIn: boolean = false;
+
   @query('#base-query-field') private baseQueryField!: HTMLInputElement;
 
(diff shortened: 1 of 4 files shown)


demo/app-root.ts
```diff
@@ -32,6 +32,8 @@ export class AppRoot extends LitElement {
 
   @state() private colGap: number = 1.7;
 
+  @state() private loggedIn: boolean = false;
```
dualcnhq, line 35: can we be consistent on naming booleans? would prefer this to be `isLoggedIn` as on some places we also use `is` for naming

dualcnhq:
just want to have the boolean naming change... other than that, this lgtm 👍

src/collection-browser.ts
```diff
@@ -984,6 +986,26 @@ export class CollectionBrowser
     const tiles: TileModel[] = [];
     docs?.forEach(doc => {
       if (!doc.identifier) return;
+
+      let loginRequired = false;
+      let contentWarning = false;
+      // Check if item and item in "modifying" collection, setting above flags
+      if (
+        doc.collections_raw?.values &&
```
iisa, line 994: hello, just some clarifying around `doc.collections_raw?.values` - is that an array or a string? line 997 looks like it is looking for an array? should it be `doc.collections_raw?.values?.length`? can/should we use the normalized, not-raw value?

src/tiles/item-image.ts
```diff
@@ -122,7 +99,9 @@ export class ItemImage extends LitElement {
 
   // Classes
   private get imageClass() {
-    return `item-image ${this.isDeemphasize ? 'deemphasize' : 'default'}`;
+    return `item-image ${
+      this.model?.contentWarning ? 'deemphasize' : 'default'
```
iisa, line 103: `contentWarning` is :nice:

iisa (commented):
looks really good, thanks for the new model fields `contentWarning` & `loginRequired`. just some clarifying questions around `collections_raw` usage from me.  is there a way to add a quick unit test around the model, component?

src/collection-browser.ts
mc2: [`collections_raw` is] The "natural" collections for an item before augmentation by side-loaded collections like ListsAPI and SimpleLists

The `collection` field above includes things like all of the fav-* collections, whereas this [`collections_raw`] is only the collections that have been directly added in the hierarchy.

mc2: `doc.collections_raw?.values` is defined as `string[] | undefined` It's either a string array or undefined.
So either: 
- truthy: string array with content
- truthy: empty string array   -- `for .. of` doesn't run
- falsy: undefined

`doc.collections_raw?.values.length` would be `number | undefined`
- truthy: number greater than 0
- falsy: 0
- falsy: undefined

Changed to `.length` as a better-ish guard -- I don't think it matters much, both ways process correctly

demo/app-root.ts
mc2: I prefer to use the `is...` prefix where it fits. The "In" in "loggedIn" serves as an implicit "is"
logged in -> true | false
"is logged in" is redundant, less clear, and not ... good English

"is required" is clearer as true | false than "required"

mc2: This is different in offshoot, because we're passing in the entire user object
https://git.archive.org/www/offshoot/-/merge_requests/216/diffs

PR #56, opened 2022-05-24
Fix resize facet jitter

mc2:
Solution summary: 
- Moved inline html into template method
- Created no-transition "preload" class
- Created toggle var to set "preload" class on loading, unset on user action

### DIFF
--- a/src/collection-browser.ts
+++ b/src/collection-browser.ts
@@ -162,6 +162,12 @@ export class CollectionBrowser
    */
   private endOfDataReached = false;
 
+  /**
+   * When the page transitions from desktop to mobile, we need to
+   * disable the transition animation.
+   */
+  private mobileFiltersloading = false;
+
   private placeholderCellTemplate = html`<collection-browser-loading-tile></collection-browser-loading-tile>`;
 
   private tileModelAtCellIndex(index: number): TileModel | undefined {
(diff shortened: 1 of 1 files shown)


src/collection-browser.ts
```diff
@@ -162,6 +162,12 @@ export class CollectionBrowser
    */
   private endOfDataReached = false;
 
+  /**
+   * When the page transitions from desktop to mobile, we need to
+   * disable the transition animation.
+   */
+  private mobileFiltersloading = false;
```
dualcnhq, line 169: suggestion: `mobileFiltersLoading`
also same question on using `is` prefix here

src/collection-browser.ts
```diff
@@ -382,6 +369,28 @@ export class CollectionBrowser
     return this.facetsLoading || this.fullYearAggregationLoading;
   }
 
+  private get mobileFacetsTemplate() {
+    return html`
+      <div id="mobile-filter-collapse">
```
iisa, line 374: idk if we want to play with the `cache` directive here? more of a question to pick your knew knowledge of built-in directives.. ^_^

src/collection-browser.ts
```diff
@@ -241,31 +247,12 @@ export class CollectionBrowser
...
-                          : ''}"
-                      >
-                        ${chevronIcon}
-                      </span>
-                      Filters
-                    </h1>
-                  </div>
-                `
-              : nothing}
+            ${this.mobileView ? this.mobileFacetsTemplate : nothing}
```
iisa, line 255: or cache at this level? idk...

src/collection-browser.ts
```diff
@@ -1101,6 +1115,18 @@ export class CollectionBrowser
       display: block;
     }
 
+    /**
+    * When page width resizes from desktop to mobile, use this class to
+    * disable expand/collapse transition when loading.
```
iisa, line 1120: 👍

src/collection-browser.ts
```diff
@@ -513,8 +522,13 @@ export class CollectionBrowser
   }
 
   handleResize(entry: ResizeObserverEntry): void {
+    const previousView = this.mobileView;
     if (entry.target === this.contentContainer) {
-      this.mobileView = entry.contentRect.width < 600;
+      this.mobileView = entry.contentRect.width < this.mobileBreakpoint;
+      // If changing from desktop to mobile disable transition
+      if (this.mobileView && !previousView) {
+        this.isResizeToMobile = true;
```
iisa, line 530: Q: when does `this.isResizeToMobile` turns to false?

iisa (commented):
great catch! had a few questions, primarily around  when `isResizeToMobile = false`

also, please may you add a unit test to check against the new `preload` class and we can see it toggle in & out during desktop/mobile view...

src/collection-browser.ts
mc2: 376, 380
```          @click=${() => {
            this.isResizeToMobile = false;
```

src/collection-browser.ts
mc2: answer above

dualcnhq (approved):
lgtm

src/collection-browser.ts
mc2: Caching of the HTML would only help if creation was coupled to blocking process, like network or DB. Since we have all the data in browser at create time, no need.

PR #60, opened 2022-05-26
Offshoot Search (Tiles): Un-bold "By" lines; remove when empty field

nsharma123:
https://webarchive.jira.com/browse/WEBDEV-5289


dualcnhq (approved):

src/tiles/grid/item-tile.ts
```diff
@@ -32,7 +32,7 @@ export class ItemTile extends LitElement {
           </div>
           <div class="item-creator">
             <div class="truncated">
-              <span><strong>By:&nbsp;</strong>${itemCreator}</span>
+              ${itemCreator ? html`<span>by&nbsp;${itemCreator}</span>` : ''}
```
iisa, line 35:
```suggestion
              ${ this.model?.creator ? html`<span>by&nbsp;${itemCreator}</span>` : nothing }
```

nothing is usually preferred.

src/tiles/grid/item-tile.ts
```diff
@@ -20,7 +20,7 @@ export class ItemTile extends LitElement {
 
   render() {
     const itemTitle = this.model?.title || '';
-    const itemCreator = this.model?.creator || '-';
+    const itemCreator = this.model?.creator || '';
```
iisa, line 23:
```suggestion
```

src/tiles/grid/item-tile.ts
iisa: please also add a quick unit test so we can keep coverage up. thank u!

mc2: Please see: https://lit.dev/docs/api/templates/#nothing

We've been using this as the right "lit.dev" way to have the DOM cleared and empty, as @iisa suggests above.

mc2, line 35: Please see: https://lit.dev/docs/api/templates/#nothing

We've been using this as the right "lit.dev" way to have the DOM cleared and empty, as @iisa suggests above.

mc2 (changes requested):

mc2 (approved):

PR #59, opened 2022-05-26
WIP Webdev 5248 logged in no preview

mc2:
Solution Summary

- Add login required icon
- Create image overlay component item-image-overlay
- Remove image overlay from item-image component

PR #67, opened 2022-06-01
WEBDEV-5248 item-overlay + loginRequired icon

dualcnhq:
**Changes:**

- Update start script and add item-overlay and loginRequired icon
- Separate these 2 files from https://github.com/internetarchive/iaux-collection-browser/pull/59
- Add loggedIn for tile-dispatching on loading icon and text overlay for deemphasized and loggedInRequired items


test/icon-overlay.test.ts
```diff
@@ -0,0 +1,30 @@
+/* eslint-disable import/no-duplicates */
+import { expect, fixture } from '@open-wc/testing';
```
mc2, line 2: Please add test for `.loggedIn=${true} .loginRequired=${true}`

test/item-image.test.ts
```diff
@@ -0,0 +1,85 @@
...
+        .baseImageUrl=${baseImageUrl}
+      >
+      </item-image>
+    `);
+
+    const dropShadow = el.shadowRoot?.querySelector('.drop-shadow');
+    // const image = dropShadow?.querySelector('img');
+
+    expect(dropShadow).to.exist;
+    /**
```
mc2, line 77: Question: should this comment be here? If so, when will we accomplish TODO?
Is this worth a Jira?

test/text-overlay.test.ts
```diff
@@ -0,0 +1,44 @@
+/* eslint-disable import/no-duplicates */
```
mc2, line 1: Please add test for `.loggedIn=${true} .loginRequired=${true}`

mc2 (approved):
@dualcnhq Please see comments and handle before merge. Thanks!

mc2 (approved):

PR #69, opened 2022-06-07
Fix waveform height

dualcnhq:
Screen Shot 2022-06-08 at 04 02 21

Screen Shot 2022-06-08 at 04 02 46

Screen Shot 2022-06-08 at 04 03 46

### DIFF
--- a/src/styles/item-image-styles.ts
+++ b/src/styles/item-image-styles.ts
@@ -7,7 +7,7 @@ export const baseItemImageStyles = css`
   .drop-shadow {
     filter: drop-shadow(1px 1px 2px rgba(0, 0, 0, 0.8));
     overflow: hidden;
-    height: 100%;
+    height: 160px;
   }
 
   .list-box {
@@ -44,8 +44,6 @@ export const waveformGradientStyles = css`
     height: 100px;
     top: 50%;
     transform: translateY(-50%);
-    /* 
-     */
   }
 
   .waveform-grad0 {
--- a/src/styles/item-image-styles.ts
+++ b/src/styles/item-image-styles.ts
@@ -40,10 +40,12 @@ export const baseItemImageStyles = css`
 export const waveformGradientStyles = css`
   .waveform {
     mix-blend-mode: screen;
-    height: 100px;
     position: relative;
+    height: 100px;
     top: 50%;
     transform: translateY(-50%);
+    /* 
+     */
   }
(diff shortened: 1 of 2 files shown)


src/tiles/item-image.ts
```diff
@@ -26,13 +26,20 @@ export class ItemImage extends LitElement {
   @query('img') private baseImage!: HTMLImageElement;
 
   render() {
+    const width = () => {
+      if (this.isListTile && !this.isCompactTile) return '100px';
```
mc2, line 30: blocker: the `.mobile` class changes the size of the `<div id="thumb">`
100px -> 90px
45px -> 30px

mc2: if `this.isListTile` the width of the waveform is always going to be `100%`

src/tiles/item-image.ts
```diff
@@ -26,13 +26,20 @@ export class ItemImage extends LitElement {
...
+    };
+
     return html`
       <div class=${classMap(this.itemBaseClass)}>
         <img
           class=${classMap(this.itemImageClass)}
           src="${this.imageSrc}"
           alt=""
           @load=${this.onLoad}
+          style="--imgWidth:${width()};"
```
mc2, line 42: question: if we just get rid of this line, won't the width just work? Or am I missing something?

src/tiles/item-image.ts
dualcnhq: fixed it

src/tiles/item-image.ts
dualcnhq: i removed it now, thanks for your feedback

src/styles/item-image-styles.ts
```diff
@@ -7,6 +7,7 @@ export const baseItemImageStyles = css`
   .drop-shadow {
     filter: drop-shadow(1px 1px 2px rgba(0, 0, 0, 0.8));
     overflow: hidden;
+    height: 160px;
```
mc2, line 10: blocker: There should not be fixed values in this component. 

Please change to `100%`

mc2 (changes requested):

src/styles/item-image-styles.ts
dualcnhq: fixed

mc2 (approved):

PR #70, opened 2022-06-07
[WEBDEV-5290] Fix stats size and spacing

dualcnhq:


src/styles/item-tile-styles.ts
```diff
@@ -0,0 +1,131 @@
+import { css } from 'lit';
```
mc2, line 1: question: Why did you remove the styles?

It would make sense to me if multiple components used the same style file. But for 1-component corresponding to 1-style file, I think I, Drini, and Isa have a preference that they be in the same component file.

mc2: blocker: If you prefer the styles in a separate file, then at least move this file into `tiles/grid`, in the same folder as the component.

Kind of a Principle: 
- The contents of a component should be together, as much as possible.

Imagine if you were distributing the item-tile component -- as a component. You would not expect the component consumer, a page author, to create a separate, global`/styles` folder.

mc2 (changes requested):

src/tiles/item-stats.ts
```diff
@@ -0,0 +1,67 @@
+import { CSSResultGroup, html, LitElement } from 'lit';
```
mc2, line 1: Great idea moving this out as a separate component.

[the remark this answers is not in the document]
mc2: blocker: We want component names to be clear and accurate.

Since this component is part of both `account-tile` and `item-tile` it's not an "item" component, but a "tile" component.

Please rename to `tile-stats.ts`

src/styles/item-stats-styles.ts
```diff
@@ -0,0 +1,40 @@
+import { css } from 'lit';
```
mc2, line 1: blocker: As below with `item-stats.ts` 

- placement next to sibling components 
- naming

Please move to `src/tiles/grid/` and rename to `tile-stats-styles.ts`

mc2 (changes requested):

src/tiles/grid/tile-stats.ts
```diff
@@ -0,0 +1,104 @@
...
+  @property({ type: Number }) viewCount?: number;
+
+  @property({ type: Number }) favCount?: number;
+
+  @property({ type: Number }) commentCount?: number;
+
+  render() {
+    console.log('tile-stats here');
+    return html`
+      <div class="stats-wrapper">
```
iisa, line 26:
```suggestion
      <div class="item-stats">
```
suggestion: create more meaningful class names

src/tiles/grid/tile-stats.ts
```diff
@@ -0,0 +1,104 @@
...
+        <div id="stats-holder">
```
iisa, line 27:
```suggestion
        <div id="list">
```

src/tiles/grid/tile-stats.ts
```diff
@@ -0,0 +1,104 @@
...
+
+  @property({ type: Number }) itemCount?: number;
+
+  @property({ type: Number }) viewCount?: number;
+
+  @property({ type: Number }) favCount?: number;
+
+  @property({ type: Number }) commentCount?: number;
+
+  render() {
```
iisa, line 23: 👍 nice & stateless

iisa (commented):
@dualcnhq good job on the style fix and locking in the desired design.

looks good!
please can you add some unit tests against the new <tile-stats> plus other elements that have been updated?
also, I have some suggestions around possibly flattening the dom even further & an open question on how we can simplify some of these class names (remove `-holder|wrapper`) and use class names that are closer to what they are responsible for...

thanks! lmk what you think

src/tiles/grid/account-tile.ts
```diff
@@ -14,43 +11,37 @@ export class AccountTile extends LitElement {
   render() {
     return html`
       <div class="outer-holder">
+        <div id="header-holder">
+          <div id="title-holder">
```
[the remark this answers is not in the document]
dualcnhq: most of this were old code when we started working on it.
just didn't touch it at the moment when i added tweaks

src/tiles/grid/tile-stats.ts
```diff
@@ -0,0 +1,104 @@
...
+    console.log('tile-stats here');
+    return html`
+      <div class="stats-wrapper">
+        <div id="stats-holder">
+          <div class="col">
+            <mediatype-icon
+              .mediatype=${this.mediatype}
+              style="--iconHeight:25px; --iconWidth:25px;"
```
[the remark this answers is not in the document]
dualcnhq: what do you mean here? @iisa

src/tiles/grid/account-tile.ts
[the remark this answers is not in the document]
iisa: Please do check if we can remove. Lesser nesting is preferred..

src/tiles/grid/tile-stats.ts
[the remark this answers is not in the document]
iisa: Move that in-line style to css declaration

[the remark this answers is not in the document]
dualcnhq: if it's about the style props, it was intentionally a passed props to mediatype-icon component as we have dynamic sizing for the mediatype-icon component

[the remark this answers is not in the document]
bfalling: @dualcnhq: Passing in sizing values is fine, but passing in using a `style` attribute is a little suspicious. e.g. If you want to pass icon height and width, then would it make more sense to use explicit parameters for height and width?

src/tiles/grid/account-tile.ts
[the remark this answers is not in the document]
nsharma123: I also want to mentioned on more thing here on class name on different tiles. there is not similarity with class name even we are displaying stats icons/count on account and item tiles.

eg. item-tiles container has different HTML structure including classes name and account-tiles does have different structure even we are showing same content like stats icon and count.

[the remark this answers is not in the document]
nsharma123: image

src/tiles/grid/account-tile.ts
```diff
@@ -14,43 +11,37 @@ export class AccountTile extends LitElement {
...
-            <div class="stat-icon">
-              ${reviewsIcon}
-              <h3>${this.model?.commentCount}</h3>
+              <span>
+                Archivist since ${this.model?.dateAdded?.getFullYear()}
+              </span>
             </div>
           </div>
         </div>
+
```
nsharma123, line 37: Thanks you !

src/tiles/grid/item-tile.ts
```diff
@@ -19,91 +16,82 @@ export class ItemTile extends LitElement {
...
   }
 
   static get styles(): CSSResultGroup {
-    const cornerRadiusCss = css`var(--collectionTileCornerRadius, 4px)`;
-
     return css`
       #container {
         background-color: #ffffff;
-        border-radius: ${cornerRadiusCss};
+        border-radius: var(--collectionTileCornerRadius, 4px);
```
nsharma123, line 62: as I remember, we don't want to use px unit in CSS instead we use rem units.

src/tiles/grid/tile-stats.ts
nsharma123: Thanks you Mark!

src/tiles/grid/tile-stats.ts
```diff
@@ -0,0 +1,104 @@
...
+            >
+            </mediatype-icon>
+          </div>
+          <div class="col">
```
nsharma123, line 35: Goody goody!

src/tiles/grid/tile-stats.ts
```diff
@@ -0,0 +1,104 @@
...
+          </div>
+        </div>
+      </div>
+    `;
+  }
+
+  static get styles(): CSSResultGroup {
+    return css`
+      .stats-wrapper {
+        height: 35px;
```
nsharma123, line 65: Open question to everyone, are we now supporting using px unit instead of rem/em units? At PB side, we totally against it unless its really needed. @iisa @mc2 @dualcnhq

src/tiles/grid/item-tile.ts
dualcnhq: it was stated in the mockups though to use `px`

src/tiles/grid/tile-stats.ts
dualcnhq: `px` was set in the mockups so i just followed it.

cc @pezvi

src/tiles/grid/item-tile.ts
iisa: @nsharma123 yep, for pixel perfect design parts like border-radius, we can use px. for text, rule of thumb is using rem.

src/tiles/grid/account-tile.ts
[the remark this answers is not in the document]
iisa: the `<tile-stats>` is a new component and looks great - encapsulated dom & style. 
the dom change is ok as it seems like we are moving to better patterns, creating scoped components, writing semantic dom ^_^

src/tiles/grid/tile-stats.ts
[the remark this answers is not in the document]
iisa: what I meant was adding this to the css block as such

```css
mediatype-icon {
  --iconHeight: 25px;
  --iconWidth: 25px;
}
```

iisa (commented):
this looks really good @dualcnhq!

in an encapsulated world like web components, we can actually be more precise with our class namings to reflect its scope.
i've got some "suggestions" added to show how we can better name the elements in `<item-tile>` and `<account-tile>`.  this really helps shape the DOM tree well as when one opens the inspector, it's clear what the responsibilities of each element is.

also, i see some nested elements with flexbox properties though their direct parent is not `display: flex` -> a sweep thru these properties to confirm whether or not we need them will help keep the components nice and tidy.

a unit test for the `<tile-stats>` display will be good to have + continue to provide "code security & peace of mind" to this repo. then adding tests to check for its existence in an item & account tile.

thanks for the good work 💪 you pared it down a lot since we talked last week.

src/tiles/grid/item-tile.ts
```diff
@@ -19,91 +16,80 @@ export class ItemTile extends LitElement {
...
+        padding-left: 5px;
+        padding-right: 5px;
+      }
+
+      .title-wrapper {
+        flex-shrink: 0;
+      }
+
+      .image-wrapper {
+        flex-grow: 1;
```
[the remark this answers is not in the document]
dualcnhq: yup got this from here: https://developer.mozilla.org/en-US/docs/Web/CSS/Layout_cookbook/Sticky_footers

src/tiles/grid/tile-stats.ts
```diff
@@ -0,0 +1,104 @@
...
+  static get styles(): CSSResultGroup {
+    return css`
+      mediatype-icon {
+        --iconHeight: 25px;
+        --iconWidth: 25px;
+      }
+
+      .item-stats {
+        height: 35px;
+        flex-shrink: 0;
```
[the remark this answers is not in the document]
dualcnhq: got the idea from this: https://developer.mozilla.org/en-US/docs/Web/CSS/Layout_cookbook/Sticky_footers

src/tiles/grid/tile-stats.ts
```diff
@@ -0,0 +1,104 @@
...
+            </p>
+          </div>
+        </div>
+      </div>
+    `;
+  }
+
+  static get styles(): CSSResultGroup {
+    return css`
+      mediatype-icon {
```
iisa, line 59: 👍

iisa (approved):
looks good, code-wise.

mc2 (approved):
Looks good to me

PR #72, opened 2022-06-08
WEBDEV-5316 - Mediatype-icon color config update

dualcnhq:
## Changes:

- update fillColor
- add quick fix for itemTile title `title` attribute display


src/tiles/mediatype-icon.ts
```diff
@@ -68,7 +68,7 @@ export class MediatypeIcon extends LitElement {
       }
 
       .fill-color {
-        fill: var(--iconCustomFillColor, var(--iconFillColor, '#000000'));
+        fill: var(--iconFillColor, '#000000');
```
mc2, line 71: Blocker: Now that the collection icon has a fill color, we can get rid of `private get collectionColor()` and `--iconCustomFillColor` in tile-list.ts and tile-list-compact.ts

Need to handle this in the same branch.

mc2 (changes requested):

src/tiles/mediatype-icon.ts
dualcnhq: updated now...

src/tiles/list/tile-list-compact.ts
```diff
@@ -45,7 +45,6 @@ export class TileListCompact extends LitElement {
           <mediatype-icon
             .mediatype=${this.model?.mediatype}
             .collections=${this.model?.collections}
-            style="--iconCustomFillColor: ${ifDefined(this.collectionColor)}"
```
mc2, line 48: Great, thanks Clau!

src/assets/img/icons/mediatype/account.ts
```diff
@@ -1,14 +1,13 @@
 import { svg } from 'lit';
 
 export const accountIcon = svg`
-  <svg viewBox="0 0 100 100"
-    xmlns="http://www.w3.org/2000/svg">
+  <svg viewBox="0 0 100 100" xmlns="http://www.w3.org/2000/svg">
+    <title>icon/account</title>
```
mc2, line 5: blocker: We don't know what the `<title>`s should be. And there are conflicting styles.

Let's not make any changes in these, but back out the changes in all the mediatype icons and wait for the right design.

There needs to be a design ticket for this. I will write it.

mc2: Let's try to center each pull request on the (hopefully) one issue of each Jira ticket. 
Here: "Item tile icon stats color"

If we do get Jira tickets are not clear and focused, let's all include in our development process: sending them back, i.e. commenting that the writer of the ticket needs to 
- be clearer
- break up the ticket
- etc.

mc2:
@dualcnhq Please note the Jira number in the first comment of every PR. In GitLab, we do it in the "description"

src/styles/item-tile-styles.ts
```diff
@@ -12,16 +12,13 @@ export const baseItemTileStyles = css`
     flex-direction: column;
```
mc2, line 12: blocker: I'm not sure if we have a coding standard for whitespace in .css definitions. But we do want whitespace between each selector and rule set. Please undo this change.

Again, `yarn format` should do this, but I'm not sure. Please `yarn format` before every commit in this project.

mc2 (changes requested):

src/styles/item-tile-styles.ts
mc2: Not resolved.

Please move file to `src/tiles/grid/` next to `item-tile.ts`

mc2 (changes requested):

mc2 (approved):

PR #73, opened 2022-06-10
Offshoot : Collection browser layout broken for rendertron (?noscript=true) 

nsharma123:
https://webarchive.jira.com/browse/WEBDEV-5266

### DIFF
--- a/package.json
+++ b/package.json
@@ -25,7 +25,7 @@
     "@internetarchive/feature-feedback": "^0.1.2",
     "@internetarchive/field-parsers": "^0.1.3",
     "@internetarchive/histogram-date-range": "^0.1.6",
-    "@internetarchive/infinite-scroller": "^0.1.2",
+    "@internetarchive/infinite-scroller": "0.1.3",
     "@internetarchive/local-cache": "^0.2.1",
     "@internetarchive/search-service": "^0.3.4",
     "@internetarchive/shared-resize-observer": "^0.2.0",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -3,7 +3,7 @@
   "description": "The Internet Archive Collection Browser.",
   "license": "AGPL-3.0-only",
   "author": "Internet Archive",
-  "version": "0.2.0",
+  "version": "0.2.2",
```
nsharma123, line 6: tag 0.2.1 was already created by clau so decided to go with 2.2

package.json
```diff
@@ -25,7 +25,7 @@
     "@internetarchive/feature-feedback": "^0.1.2",
     "@internetarchive/field-parsers": "^0.1.3",
     "@internetarchive/histogram-date-range": "^0.1.6",
-    "@internetarchive/infinite-scroller": "^0.1.2",
+    "@internetarchive/infinite-scroller": "0.1.3",
```
mc2, line 28: Suggest "^0.1.3"

About using the caret ^: https://nodesource.com/blog/semver-tilde-and-caret/

mc2 (changes requested):

package.json
dualcnhq: @nsharma123 is there a specific reason why you removed the caret here?

i also just learned this versioning, more info on source [here](https://javascript.plainenglish.io/the-math-behind-npm-versioning-5236f8c821c2)

image

dualcnhq:
would recommend to create a separate PR commit for the version bump to 0.2.2 so we have clear release tag changelog

dualcnhq (approved):

PR #78, opened 2022-06-22
WEBDEV-5332 - Implement semantic HTML and add screen-reader only class

dualcnhq:
Reference: https://learn-the-web.algonquindesign.ca/topics/html-semantics-cheat-sheet/

### DIFF
--- a/src/tiles/grid/tile-stats.ts
+++ b/src/tiles/grid/tile-stats.ts
@@ -24,17 +24,18 @@ export class TileStats extends LitElement {
     return html`
       <div class="item-stats">
         <p class="sr-only">
-          ${this.mediatype === 'account' ? 'Account' : 'Item'} Stats:
+          ${this.mediatype === 'account' ? 'Account Stats' : 'Item Stats'}
         </p>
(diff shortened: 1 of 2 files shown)


src/tiles/grid/tile-stats.ts
```diff
@@ -23,33 +23,41 @@ export class TileStats extends LitElement {
...
+              <span class="sr-only">Favorites count:</span>
               ${formatCount(this.favCount, 'short', 'short')}
             </p>
-          </div>
-          <div class="col">
+          </li>
+          <li class="col">
             ${reviewsIcon}
             <p class="status-text">
+              <span class="sr-only">Reviews count:</span>
```
iisa, line 56:
```suggestion
              <span class="sr-only">Reviews:</span>
```

iisa (commented):
looks good. just the next batch of clean up before your unit tests.
- update labels
- keep sr-only tags in tile-stats
-

iisa (approved):
so much nicer and much more readable! :lgtm: 🎉

PR #83, opened 2022-06-24
Webdev 5312 replace by text with sorted date

nsharma123:
https://webarchive.jira.com/browse/WEBDEV-5312

### DIFF
--- a/src/tiles/grid/item-tile.ts
+++ b/src/tiles/grid/item-tile.ts
@@ -4,6 +4,9 @@ import { customElement, property } from 'lit/decorators.js';
 
 import { TileModel } from '../../models';
 import { formatCount } from '../../utils/format-count';
+import { formatDate } from '../../utils/format-date';
+
+import { SortParam } from '@internetarchive/search-service';
 
 import { favoriteFilledIcon } from './icons/favorite-filled';
 import { reviewsIcon } from './icons/reviews';
@@ -18,9 +21,33 @@ export class ItemTile extends LitElement {
 
   @property({ type: String }) baseImageUrl?: string;
 
+  @property({ type: Object }) sortParam: SortParam | null = null;
+
+  private get getSortedValue() {
+    let sortedValue = '' as any;
+
+    switch (this.sortParam?.field) {
+      case 'date':
+        sortedValue = { field : 'published', value: this.model?.datePublished };
+      case 'reviewdate':
+        sortedValue = { field : 'reviewed', value: this.model?.dateReviewed };
+      case 'addeddate':
+        sortedValue = { field : 'added', value: this.model?.dateAdded };
+      case 'publicdate':
+        sortedValue = { field : 'archived', value: this.model?.dateArchived };
+
+      return html`<span>${sortedValue.field as string} ${formatDate(sortedValue.value, 'long')}</span>`;
+    }
+  }
+
+  private get getCreator() {
+    return this.model?.creator
+      ? html`<span>by&nbsp;${this.model?.creator}</span>`
+      : nothing;
+  }
+
   render() {
     const itemTitle = this.model?.title || nothing;
-    const itemCreator = this.model?.creator;
     return html`
       <div id="container">
         <div id="title-image-container">
@@ -31,9 +58,9 @@ export class ItemTile extends LitElement {
           </div>
           <div class="item-creator">
             <div class="truncated">
-              ${itemCreator
-                ? html`<span>by&nbsp;${itemCreator}</span>`
-                : nothing}
+              ${this.sortParam 
+                ? this.getSortedValue
+                : this.getCreator}
             </div>
           </div>
         </div>
--- a/src/tiles/tile-dispatcher.ts
+++ b/src/tiles/tile-dispatcher.ts
@@ -154,6 +154,7 @@ export class TileDispatcher
               .currentHeight=${this.currentHeight}
               .collectionNameCache=${this.collectionNameCache}
               .baseImageUrl=${this.baseImageUrl}
+              .sortParam=${sortParam}
             ></item-tile>`;
         }
       case 'list-compact':
--- a/package.json
+++ b/package.json
@@ -98,9 +98,6 @@
   },
   "lint-staged": {
     "*.ts": [
-      "eslint --fix",
-      "prettier --write",
-      "git add"
     ]
   }
 }


src/tiles/grid/item-tile.ts
```diff
@@ -34,7 +38,11 @@ export class ItemTile extends LitElement {
             </item-image>
           </div>
           <div class="created-by truncated">
-            ${itemCreator ? html`<span>by&nbsp;${itemCreator}</span>` : nothing}
+            ${
+              this.sortParam
```
iisa, line 42: suggestion: make this a getter so that you can make a unit test against it

src/tiles/grid/item-tile.ts
```diff
@@ -15,9 +17,11 @@ export class ItemTile extends LitElement {
 
   @property({ type: String }) baseImageUrl?: string;
 
+  @property({ type: Object }) sortParam: SortParam | null = null;
```
iisa, line 20:
```suggestion
  @property({ type: Object }) sortParam?: SortParam;
```
denotes that it is optional and can default to undefined

src/tiles/grid/item-tile.ts
iisa: Q: when does `this.sortParam` return falsy?

src/tiles/grid/item-tile.ts
```diff
@@ -49,6 +57,38 @@ export class ItemTile extends LitElement {
     `;
   }
 
+  private get sortedDateInfoTemplate() {
+    let sortedValue = {} as any;
+
+    switch (this.sortParam?.field) {
```
iisa, line 63: perhaps we use 1 function to return to return instead of the ternary in the render?

```suggestion
        if (!this.sortParam) {
          return this.creatorTemplate;
        }
    switch (this.sortParam?.field) {
```
I only note because you have `return this.creatorTemplate;` at the end of switch case

iisa: good to have unit test for each permutation to verify what renders.

src/tiles/tile-dispatcher.ts
```diff
@@ -154,6 +154,7 @@ export class TileDispatcher
               .currentHeight=${this.currentHeight}
               .collectionNameCache=${this.collectionNameCache}
               .baseImageUrl=${this.baseImageUrl}
+              .sortParam=${sortParam}
```
iisa, line 157: Q: when does this every return null?

iisa (commented):
let's get some unit tests in for each permutation

src/tiles/grid/item-tile.ts
nsharma123: I thought around early return but `this.sortParam?.field` can also have `titleSorter`, `titleSorter` values which is not relevant to current task and in these cases, we need to return createTemplate. that's why if sort field if not among these 4 case:, just return createTemplate as default.

Will check unit tests for these pieces.

src/tiles/grid/item-tile.ts
```diff
@@ -49,6 +57,38 @@ export class ItemTile extends LitElement {
...
+        sortedValue = { field: 'added', value: this.model?.dateAdded };
+        break;
+      case 'publicdate':
+        sortedValue = { field: 'archived', value: this.model?.dateArchived };
+        break;
+      default:
+        return this.creatorTemplate;
+    }
+
+    return html`<span
```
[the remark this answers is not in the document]
nsharma123: I always relay on yarn format command and sometimes its makes code formatting too weird :) Changing per your thoughts. Thanks!

src/tiles/tile-dispatcher.ts
nsharma123: So if you don't have any ?sort=xxx param in URL, it does return null value.

src/tiles/grid/item-tile.ts
nsharma123: when you don't have any ?sort param in URL..

src/tiles/grid/item-tile.ts
```diff
@@ -34,7 +38,11 @@ export class ItemTile extends LitElement {
             </item-image>
           </div>
           <div class="created-by truncated">
-            ${itemCreator ? html`<span>by&nbsp;${itemCreator}</span>` : nothing}
+            ${
+              this.sortParam
+                ? this.sortedDateInfoTemplate
+                : this.creatorTemplate
```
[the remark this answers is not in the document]
nsharma123: If there is not ?sort param in URL, we can directly call `creatorTemplate`.

src/tiles/grid/item-tile.ts
```diff
@@ -49,6 +57,38 @@ export class ItemTile extends LitElement {
     `;
   }
 
+  private get sortedDateInfoTemplate() {
+    let sortedValue = {} as any;
+
+    switch (this.sortParam?.field) {
+      case 'date':
+        sortedValue = { field: 'published', value: this.model?.datePublished };
```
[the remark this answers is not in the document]
nsharma123: its different thing, that's why I need to build code snippet to make field/value object here. eg. sortParam.field value is `reviewdate`, we want to show `reviewed` (an different text label)

src/tiles/grid/item-tile.ts
```diff
@@ -49,6 +57,38 @@ export class ItemTile extends LitElement {
...
+        break;
+      case 'publicdate':
+        sortedValue = { field: 'archived', value: this.model?.dateArchived };
+        break;
+      default:
+        return this.creatorTemplate;
+    }
+
+    return html`<span
+      >${sortedValue.field as string}
```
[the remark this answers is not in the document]
nsharma123: I thought but models.ts file already have lots of data related to sort and date variables also this function used `sortedValue` inside this only. so I think we should increase more type for sort&date fields atleast.

src/tiles/grid/item-tile.ts
```diff
@@ -49,6 +54,51 @@ export class ItemTile extends LitElement {
...
+        break;
+      case 'publicdate':
+        sortedValue = { field: 'archived', value: this.model?.dateArchived };
+        break;
+      default:
+        break;
+    }
+
+    return html`
+      <div class="truncated date-sorted-by">
```
dualcnhq, line 84: a little nitpick i guess, can we place the className as `date-sorted-by` goes first followed by `truncated`?
doing so, it will be similar to the other parts where we add `truncated` className

test/tiles/grid/item-tile.test.ts
```diff
@@ -0,0 +1,116 @@
+/* eslint-disable import/no-duplicates */
```
dualcnhq, line 1: great to have tests now for this component 👍 💪

PR #94, opened 2022-08-05
Dompurify add vendor type

iisa:
problem: offshoot is not properly building bc the vendor package `dompurify` has an implicit type.

solution: make shim type file to declare it

acceptance: offshoot cleanly builds

note: 0.2.14-a1

### DIFF
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -17,5 +17,5 @@
     "declaration": true,
     "importsNotUsedAsValues": "error"
   },
-  "include": ["src", "test", "index.ts"],
+  "include": ["src", "test", "index.ts", "types"],
 }
--- a/types/dompurify.d.ts
+++ b/types/dompurify.d.ts
@@ -1,0 +1,1 @@
+declare module 'dompurify';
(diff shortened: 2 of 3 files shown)


mmcnl:
@iisa I don't understand the backstory on this but LGTM. I do notice that the codecov check failed (https://github.com/internetarchive/iaux-collection-browser/pull/94/checks?check_run_id=7698782961)

iisa:
hi @mmcnl thanks for the quick CR.

re: coverage drop, I guess adding type context opened up lines to check? here's the diff - nothing I touched https://app.codecov.io/gh/internetarchive/iaux-collection-browser/compare/94/changes

re backstory:
offshoot apparently type checks all ts files that it uses to build and it caught `Dompurify` as `any` - blocking @dualcnhq 's latest offshoot MR https://git.archive.org/www/offshoot/-/merge_requests/237
image

so, this is direct fix for it. and [our work around for this](https://github.com/cure53/DOMPurify/issues/705) 

I will take the test hit bc it is out of scope for my debug time and we can work to make up for it in the next rounds. hope that's ok with you.

mmcnl:
@iisa sure, totally makes sense, thanks for filling me in on your process. I just would want us to not get in a long-term habit of ignoring codecov (or other) CI checks

PR #98, opened 2022-08-17
WEBDEV-5340 WEBDEV-5341 Text snippet blocks for FTS

latonv:
Adds text snippet blocks to collection browser tiles and list items to display keyword matches for full-text searches.


src/tiles/list/tile-list.ts
```diff
@@ -292,6 +292,19 @@ export class TileList extends LitElement {
     );
   }
 
+  private get textSnippetsTemplate() {
+    if (!this.hasSnippets) return nothing;
+
+    return html`
+      <text-snippet-block .viewSize=${'list'}></text-snippet-block>
```
iisa, line 299:
```suggestion
      <text-snippet-block viewSize="list"></text-snippet-block>
```
I want to say writing it as attribute removes the linter's bark this need to interpolate `=${'list'}`

src/tiles/text-snippet-block.ts
```diff
@@ -0,0 +1,116 @@
...
+import { customElement, property } from 'lit/decorators.js';
+import { unsafeHTML } from 'lit/directives/unsafe-html.js';
+import { map } from 'lit/directives/map.js';
+import { join } from 'lit/directives/join.js';
+import DOMPurify from 'dompurify';
+
+@customElement('text-snippet-block')
+export class TextSnippetBlock extends LitElement {
+  // The default snippets value here is just for testing's sake; will be removed.
+  @property({ type: Array }) snippets?: string[] = [
```
iisa, line 11:
```suggestion
  @property({ type: Array }) snippets?: string[] = [];
```
hmmm, what's a good way to toggle a set of results with FTS snippets from `<app-root>` so we may remove this placeholder?

light digging, looks like we can probably create an FTS SearchService stub that returns a hard coded set of results.  So that when toggling a checkbox/radio button on demo controls, we can just display the results set with `snippet` included in - like in model update.

what do you think?

src/tiles/image-block.ts
```diff
@@ -94,6 +97,11 @@ export class ImageBlock extends LitElement {
         flex: 1;
       }
 
+      .grid.has-snippets {
+        /* If there is a text snippet block present, the image block needs to shrink */
+        height: 11rem;
```
iisa, line 102: Suggestion: remove image block's awareness of "snippets" to keep component scope down.

we can probably control this height from the tile level? perhaps... toggle a CSS if tile has snippets by setting something like `style="var(--imgHeight: 11rem;)" whenever it needs it.

so in `.grid { }` rules, we can set default and entry like:
```
.grid { height: var(--imgBlockHeight,  16rem) };

File: src/tiles/grid/item-tile.ts [hunk 4]
```diff
@@ -40,9 +41,12 @@ export class ItemTile extends LitElement {
             .loggedIn=${this.loggedIn}
             .isCompactTile=${false}
             .isListTile=${false}
+            .hasSnippets=${this.hasSnippets}
```
iisa, line 44:
```suggestion
            style=${this.hasSnippets? '--imageBlockHeight: 11rem;' : ''}
```
something like this.. see my comment in item-tile about component scope.

src/tiles/list/tile-list.ts
```diff
@@ -292,6 +292,19 @@ export class TileList extends LitElement {
     );
   }
 
+  private get textSnippetsTemplate() {
```
iisa, line 295: suggestion: add return type

src/tiles/text-snippet-block.ts
```diff
@@ -0,0 +1,116 @@
...
+   */
+  private get ellipsisJoinedSnippets() {
+    const sanitizeOptions = { ALLOWED_TAGS: ['mark'] };
+
+    return join(
+      map(
+        this.snippets,
+        s => html`
+          <span>
+            ${unsafeHTML(
```
iisa, line 46: ahh..

src/tiles/text-snippet-block.ts
```diff
@@ -0,0 +1,116 @@
...
+              DOMPurify.sanitize(
+                TextSnippetBlock.markMatches(s),
```
iisa, line 48: Q: any particular reason why we need/want this fn to be static?

src/tiles/text-snippet-block.ts
```diff
@@ -0,0 +1,116 @@
...
+   * in many screen-readers (though there are ways to work around this if needed).
+   */
+  private static markMatches(snippet: string) {
+    return snippet.replace(/{{{(.+?)}}}/g, '<mark>$1</mark>');
+  }
+
+  static get styles(): CSSResultGroup {
+    return css`
+      #container {
+        display: -webkit-box;
```
iisa, line 76: Q: curious on this choice?

iisa (commented):
looks really good!
got 2 suggestions to check on so that we can actually move this code into the mainline independent of the service work.
- moving stub into demo (toggling static result set w/ checkbox)
- removing `image-block`'s awareness of snippets
- + unit tests

exciting times

src/tiles/text-snippet-block.ts
latonv: Thanks Isa! Yeah, I was grappling with what might be a reasonable way to apply the test snippets for the demo app. This approach sounds good to me.

src/tiles/list/tile-list.ts
latonv: 👍

src/tiles/image-block.ts
latonv: That makes more sense yeah -- I'll give this a try.

src/tiles/text-snippet-block.ts
latonv: I believe `display: -webkit-box` is required for the line-clamp rule below to work correctly.

src/tiles/list/tile-list.ts
latonv: 👍

src/tiles/text-snippet-block.ts
latonv: My thinking was that it's an instance-independent util fn and probably not useful outside of this class (though I may be wrong about that!).

latonv: Refactored this into a getter that produces marked versions of a component's snippets.

src/tiles/text-snippet-block.ts
latonv: The snippets can now be toggled from the demo controls (by wrapping the search-service to add hard-coded snippets). The placeholders are no longer present.

src/tiles/text-snippet-block.ts
latonv: Latest version no longer relies on unsafeHTML or DOMPurify.

src/tiles/grid/item-tile.ts
```diff
@@ -109,6 +120,19 @@ export class ItemTile extends LitElement {
...
+    if (!this.hasSnippets) return nothing;
+
+    return html`<text-snippet-block
+      viewsize="grid"
+      .snippets=${this.model?.snippets}
+    ></text-snippet-block>`;
+  }
+
+  private get hasSnippets(): boolean {
+    return !!(this.model?.snippets && this.model.snippets.length > 0);
```
iisa, line 133: can we do `this.model?.snippets?.length` ?

src/tiles/grid/item-tile.ts
```diff
@@ -142,14 +166,23 @@ export class ItemTile extends LitElement {
...
       }
 
+      image-block {
+        display: block;
+        margin-bottom: 5px;
+      }
+
+      image-block.has-snippets {
+        /* If there is a text snippet block present, the image block needs to shrink */
+        --imgBlockHeight: 11rem;
```
iisa, line 176: :nice:

src/tiles/list/tile-list.ts
```diff
@@ -140,7 +140,7 @@ export class TileList extends LitElement {
         ${this.viewsTemplate} ${this.ratingTemplate} ${this.reviewsTemplate}
       </div>
       ${this.topicsTemplate} ${this.collectionsTemplate}
-      ${this.descriptionTemplate}
+      ${this.descriptionTemplate} ${this.textSnippetsTemplate}
```
iisa, line 143: suggestion: add unit tests to cover this method and call

src/tiles/text-snippet-block.ts
```diff
@@ -0,0 +1,126 @@
...
+    `;
+  }
+
+  /**
+   * An array of HTML templates derived from the snippets, with ellipses inserted
+   * at the beginning, end, and between each pair of snippets.
+   */
+  private get ellipsisJoinedSnippets(): TemplateResult {
+    return html`&hellip; ${join(this.snippetTemplates, html` &hellip; `)}
+    &hellip;`;
```
iisa, line 38: q: was the line break because of max-line length rule? perhaps we move the whole `html` block under the return so it can be 1 line..

src/tiles/text-snippet-block.ts
```diff
@@ -0,0 +1,126 @@
...
+   *
+   * Note on `<em>` vs. `<mark>`:
+   * The old search page snippets had search keywords demarcated with `<em>` tags.
+   * The `<mark>` tag is semantically more accurate for this use case than `<em>`,
+   * but screen-reader behavior may be different. `<em>` will likely be read in a
+   * different tone, while `<mark>` is often read no differently than ordinary text
+   * in many screen-readers (though there are ways to work around this if needed).
+   */
+  private get snippetTemplates(): TemplateResult[] | undefined {
+    return this.snippets?.map(s => {
```
iisa, line 56: great isolation! letting the class.component handle this itself 👍 

also, nice on the render early return so this doesn't run unless it needs to 👍

src/tiles/text-snippet-block.ts
```diff
@@ -0,0 +1,126 @@
...
+      return html`<span>${templates}</span>`;
+    });
+  }
+
+  static get styles(): CSSResultGroup {
+    return css`
+      #container {
+        display: -webkit-box;
+        font-family: 'Times New Roman', serif;
+        color: #2c2c2c;
```
iisa, line 88: you _probably_ don't need `color` as it is a global css var that cascades into Shadow DOM...
Q: does this text color always match the tile's main color? if so, you may not need to set ^_^ good to double check.

src/tiles/text-snippet-block.ts
```diff
@@ -0,0 +1,126 @@
...
+
+      // Squash everything into a single span template
+      return html`<span>${templates}</span>`;
+    });
+  }
+
+  static get styles(): CSSResultGroup {
+    return css`
+      #container {
+        display: -webkit-box;
```
iisa, line 86: Q: do we also need to support other display types here?

src/tiles/text-snippet-block.ts
```diff
@@ -0,0 +1,126 @@
...
+@customElement('text-snippet-block')
+export class TextSnippetBlock extends LitElement {
+  @property({ type: Array }) snippets?: string[] = [];
+
+  @property({ type: String }) viewSize: string = 'desktop';
+
+  render() {
+    const viewSizeClass = this.viewSize === 'grid' ? 'grid' : 'list';
+
+    if (!this.snippets || this.snippets.length === 0) return html`${nothing}`;
```
iisa, line 21: 👍 :yesgood:

(does `!this.snippets?.length` work?)

test/text-snippet-block.test.ts
```diff
@@ -0,0 +1,69 @@
...
+import '../src/tiles/text-snippet-block';
+
+describe('TextSnippetBlock component', () => {
+  it('should render initial component', async () => {
+    const el = await fixture<TextSnippetBlock>(
+      html`<text-snippet-block></text-snippet-block>`
+    );
+
+    // No content if no snippets
+    expect(el.shadowRoot?.children.length).to.equal(0);
```
iisa, line 14: ✔️

test/text-snippet-block.test.ts
```diff
@@ -0,0 +1,69 @@
...
+      ></text-snippet-block>`
+    );
+
+    const container = el.shadowRoot?.querySelector('#container');
+
+    // Applies the right container classes
+    expect(container?.classList.contains('list')).to.be.true;
+    expect(container?.classList.contains('grid')).to.be.false;
+  });
+});
```
iisa, line 69: 👏 thanks for tests

test/tiles/grid/item-tile.test.ts
```diff
@@ -113,4 +113,23 @@ describe('Item Tile', () => {
...
+      </item-tile>
+    `);
+
+    const snippetBlock = el.shadowRoot?.querySelector('text-snippet-block');
+
+    expect(snippetBlock).to.exist;
+  });
+
+  it('should not render snippet block when no snippets are present', async () => {
+    const el = await fixture<ItemTile>(html`<item-tile></item-tile>`);
```
iisa, line 129: 👍

iisa (approved):
looks great! thank u for addressing the `unsafeHTML` issue by having the element class generate it when needed + the defensive way of adding it onto the tile list views + unit tests 

small suggestions, comments, questions, but otherwise lgtm once the dynamic demo checkbox is back in play ^_^

src/tiles/grid/item-tile.ts
latonv: 👍

src/tiles/text-snippet-block.ts
latonv: Yeah, I think this was Prettier's doing. Will fix the formatting here.

src/app-root.ts
```diff
@@ -206,6 +214,46 @@ export class AppRoot extends LitElement {
...
+    } else {
+      // Restore the default seach service
+      this.searchService = SearchService.default;
+    }
+
+    // Re-perform the current search to show/hide the snippets immediately
+    const oldQuery = this.searchQuery;
+    this.searchQuery = ''; // Should just reset to the placeholder
+    await this.updateComplete;
+    this.searchQuery = oldQuery; // Re-apply the original query
```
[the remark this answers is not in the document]
latonv: This doesn't seem to work in Safari, but I'm a bit baffled as to why. Chrome/FF/Edge all appear to handle it fine (at least for me). Not clear what Safari has an issue with here.

src/tiles/text-snippet-block.ts
latonv: The `-webkit-line-clamp` approach to ellipsizing the snippet block requires `display: -webkit-box` (and notably, this is widely cross-browser compatible despite the prefix). That said, there are definitely tradeoffs here.

A minor variation from our design spec in this CSS-only approach is that we can't guarantee there will be a space before the end ellipsis. There are some hacky CSS ways around this by abusing float behavior, but those come with their own set of drawbacks. We could also just perform custom truncation via a script and accept the added complexity/performance cost of that.

I suppose if we want the outer container element to have a customizable display type, we could just have a separate inner container that enforces the -webkit-box ellipsis stuff. But my feeling is that we wouldn't really gain much that we can't already do by just setting the display of the host element.

src/tiles/text-snippet-block.ts
latonv: 👍

src/tiles/text-snippet-block.ts
latonv: Yep, looks like this is the same as the inherited value so not needed! 👍

src/app-root.ts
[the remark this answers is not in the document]
latonv: Should be working now -- Safari doesn't handle the second searchQuery assignment properly if it occurs in the same tick as the previous update. So a 0ms timeout trick fixes it. Not ideal, but should be fine for the demo app.

src/tiles/list/tile-list.ts
latonv: List items had no unit tests at all to begin with, so I've added some to cover this and other basic list stuff.

iisa (approved):
looks good! demo is behaving well also 🚀

dualcnhq (approved):

PR #101, opened 2022-08-24
Add analytics

dualcnhq:


src/analytics/analytics-handler-interface.ts
```diff
@@ -0,0 +1,69 @@
...
+
+  /**
+   * Send a sampled event
+   *
+   * @param {options} AnalyticsEvent
+   */
+  sendEvent(options: AnalyticsEvent): void;
+
+  /** @deprecated use sendEvent instead */
+  send_event(
```
latonv, line 20: Q: Curious about the need for this deprecated method. Does collection-browser have subcomponents that still expect this?

src/collection-browser.ts
```diff
@@ -368,24 +372,44 @@ export class CollectionBrowser
...
   }
 
   private displayModeChanged(
     e: CustomEvent<{ displayMode: CollectionDisplayMode }>
   ) {
     this.displayMode = e.detail.displayMode;
+
+    this.analyticsHandler?.sendEvent({
+      category: 'collection-browser',
+      action: `change viewMode: ${this.displayMode}`
```
latonv, line 389: Could this action be `viewModeChanged: ${this.displayMode}` for consistency with the other events? (Not sure if these actions are existing tags we're already using 😄)

src/collection-browser.ts
```diff
@@ -155,6 +157,8 @@ export class CollectionBrowser
 
   private languageCodeHandler = new LanguageCodeHandler();
 
+  private analyticsHandler?: AnalyticsHandlerInterface;
```
latonv, line 160: Q: Would this make sense as a reactive property rather than a private field? Does Lit allow this to be set the same way as other properties?

src/collection-facets.ts
```diff
@@ -117,6 +120,11 @@ export class CollectionFacets extends LitElement {
       detail: this.selectedFacets,
     });
     this.dispatchEvent(event);
+
+    this.analyticsHandler?.sendEvent({
+      category: 'collection-browser',
+      action: 'facetsChanged'
```
latonv, line 126: Q: Should this also send which facet was changed?

latonv (approved):
👍 LGTM!

Just had a few small questions, no blockers. Should be good to go once you've added the tests!

Note on the merge conflict: I added the collection-name-cache alpha version in my PR #98, but will be updating this to 0.1.5 proper.

src/analytics/analytics-handler-interface.ts
dualcnhq: i just copy pasta this from class from offshoot... we can probably remove it here then

nsharma123:
I would prefer to create a config file to store all the eventCategory and actions and then call the as particular files. something like this https://github.com/internetarchive/iaux-book-actions/blob/main/src/core/config/analytics-event-and-category.js

src/app-root.ts
```diff
@@ -179,6 +190,7 @@ export class AppRoot extends LitElement {
           .collectionNameCache=${this.collectionNameCache}
           .showHistogramDatePicker=${true}
           .loggedIn=${this.loggedIn}
+          .analyticsHandler=${this.analyticsHandler}
```
iisa, line 193: we will make our own mock handler to add to view to show latest event captured

src/collection-browser.ts
iisa: yes, should be property 👍  - handy housekeeping thing to do is put `attribute: false` as lit  defaults to true @property({ type: Object, attribute: false })

src/collection-browser.ts
```diff
@@ -368,24 +372,44 @@ export class CollectionBrowser
     const sortField = SortFieldToMetadataField[this.selectedSort];
     if (!sortField) return;
     this.sortParam = { field: sortField, direction: this.sortDirection };
+
+    this.analyticsHandler?.sendEvent({
+      category: 'collection-browser',
+      action: `selectedSortChanged - sortParam: ${this.sortParam}`,
```
iisa, line 378:
```
category: 'collection-browser',
action: `sortBy`,  Q: defaults? yes.  - let's clarify what the specific sort is
label: `${sortField}-${direction}`
```

src/collection-browser.ts
iisa, line 389:
```
category: 'collection-browser',
action: 'displayMode'
label: this.dm
```

src/collection-browser.ts
```diff
@@ -368,24 +372,44 @@ export class CollectionBrowser
...
   }
 
   private selectedTitleLetterChanged() {
     this.titleQuery = this.selectedTitleFilter
       ? `firstTitle:${this.selectedTitleFilter}`
       : undefined;
+
+    this.analyticsHandler?.sendEvent({
+      category: 'collection-browser',
+      action: `selectedTitleLetterChanged: ${this.titleQuery}`
```
iisa, line 400:
```
category: 'collection-browser',
action: `sortByTitle`
label: this.letter
```

src/collection-browser.ts
```diff
@@ -368,24 +372,44 @@ export class CollectionBrowser
...
   }
 
   private selectedCreatorLetterChanged() {
     this.creatorQuery = this.selectedCreatorFilter
       ? `firstCreator:${this.selectedCreatorFilter}`
       : undefined;
+
+    this.analyticsHandler?.sendEvent({
+      category: 'collection-browser',
+      action: `selectedCreatorLetterChanged: ${this.creatorQuery}`
```
iisa, line 411:
```
category: 'collection-browser',
       action: sortByCreator
label: letter

```

src/collection-facets.ts
iisa, line 126: category: 'collection-browser',
action: 'facetsChanged'
label: facet

iisa (commented):
great first push and thanks for side-by-side code walk thru.

✔️ got good convention with no_sampling + dynamic label value


will add:
- [x] demo entry to show latest event fired
- [x] use npm package
- [x] unit test
- [x] a sprinkle more of tests?
- [ ] 🍰 

thank u for great first push

side effect: yay we get more standard in ts template :yesgood:

thanks @dualcnhq

src/analytics/analytics-handler-interface.ts
dualcnhq: removing this file

src/analytics/analytics-handler-interface.ts
```diff
@@ -0,0 +1,69 @@
+import type { AnalyticsEvent } from '@internetarchive/analytics-manager';
+
+export interface AnalyticsHandlerInterface {
```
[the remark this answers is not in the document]
dualcnhq: yup, removing this file

src/collection-browser.ts
```diff
@@ -226,6 +236,10 @@ export class CollectionBrowser
   @query('infinite-scroller')
   private infiniteScroller!: InfiniteScroller;
 
+  private analyticsCategories = analyticsCategories;
```
iisa, line 239: hmmm, thinking about usage. _who_ gets to declare the category - the component or the page the component is on? 
I like having a default like `collection-browser` but I think it should get set as a lit reactive property to receive the magic when we drop this component into the new collection page, or when its on a favorites page, etc.

suggestion: move to reactive lit property with default as `collection-browser`
this:
  - removes `analyticsCategories` object/enum
  - since the new property can be set with intention at the top-level, its name can be less generic, too.  like, "searchContext"
  - adds new property in demo like, `. searchContext=${'in-repo-demo'}`

```suggestion
```

src/collection-browser.ts
```diff
@@ -368,24 +385,48 @@ export class CollectionBrowser
     const sortField = SortFieldToMetadataField[this.selectedSort];
     if (!sortField) return;
     this.sortParam = { field: sortField, direction: this.sortDirection };
+
+    this.analyticsHandler?.sendEventNoSampling({
+      category: this.analyticsCategories.default,
+      action: this.analyticsActions.sortBy,
+      label: `${sortField} - ${this.sortDirection}`,
```
iisa, line 392: suggestion: remove white space from `-` for **all** `label` values

src/collection-browser.ts
```diff
@@ -368,24 +385,48 @@ export class CollectionBrowser
...
+    });
   }
 
   private displayModeChanged(
     e: CustomEvent<{ displayMode: CollectionDisplayMode }>
   ) {
     this.displayMode = e.detail.displayMode;
+
+    this.analyticsHandler?.sendEventNoSampling({
+      category: this.analyticsCategories.default,
```
iisa, line 402: so with the removal of `analyticsCategories` enum, we can write `category: ${this.searchContext}` - much more descriptive 👍

src/collection-browser.ts
```diff
@@ -368,24 +385,48 @@ export class CollectionBrowser
...
 
   private selectedTitleLetterChanged() {
     this.titleQuery = this.selectedTitleFilter
       ? `firstTitle:${this.selectedTitleFilter}`
       : undefined;
+
+    this.analyticsHandler?.sendEventNoSampling({
+      category: this.analyticsCategories.default,
+      action: this.analyticsActions.sortByTitle,
+      label: `${this.titleQuery}`,
```
iisa, line 416:
```suggestion
      label: this.titleQuery ?? 'unset',
```
sometimes the label values have string interpolation, some none...please do take a sweep to check strip unnecessary transforms, thanks ^_^

src/collection-browser.ts
```diff
@@ -368,24 +385,48 @@ export class CollectionBrowser
...
 
   private selectedCreatorLetterChanged() {
     this.creatorQuery = this.selectedCreatorFilter
       ? `firstCreator:${this.selectedCreatorFilter}`
       : undefined;
+
+    this.analyticsHandler?.sendEventNoSampling({
+      category: this.analyticsCategories.default,
+      action: this.analyticsActions.sortByCreator,
+      label: `${this.creatorQuery}`,
```
iisa, line 428: note above

src/analytics/analytics-handler-interface.ts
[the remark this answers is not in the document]
iisa: didn't see this being used during my last pass - im quite sure this is already provided in the npm pkg...

src/app-root.ts
```diff
@@ -34,12 +41,28 @@ export class AppRoot extends LitElement {
...
+  private analyticsHandler: AnalyticsHandlerInterface = {
+    sendPing: this.sendAnalytics,
+    sendEvent: this.sendAnalytics,
+    sendEventNoSampling: this.sendAnalytics,
+  };
+
+  private sendAnalytics(ae: AnalyticsEvent) {
+    this.actionBlob = `{category: ${ae.category}, action: ${ae.action}, label: ${ae.label}}`;
+    console.log('actionBlob: ', this.actionBlob);
+    this.analyticsManager?.sendEventNoSampling(ae);
```
iisa, line 63: Q: does this loop?

iisa (changes requested):
hi clau! this is getting better and better.

Q: is the tile / result click in this PR as well? I didn't see it

A few notes:
- keep analytics calls at `<collection-browser>` level. `<collection-facets>` can return a callback property when facets change, so that it does not need to "know" any analytics.
- removing `analyticsCategories` enum to prefer for a more declarative `searchContext` lit property so that the pages that use the component can better define.
- trimming white space from `label` field
- various notes around label value, esp. facetsChanged + sending `'unset'` instead of `'undefined'`

at this checkpoint, these changes shouldn't be too hard to implement.  That way you are clear for unit tests. 👍 

thanks and looking forward to final pass thru! ☺️

PR #107, opened 2022-09-06
WEBDEV-5450 Convert collection facet titles to links

latonv:
Renders collection facet titles as links to their collection page.

Note that the links are domain-relative, so on the demo app they link to non-existent pages. They should be fine on archive.org though.

### DIFF
--- a/src/collection-facets.ts
+++ b/src/collection-facets.ts
@@ -352,7 +352,11 @@ export class CollectionFacets extends LitElement {
               facetGroup.key !== 'collection'
                 ? html`${bucket.displayText ?? bucket.key}`
                 : html`
-                    <a href="/details/${bucket.key}" title="">
+                    <a
+                      href="/details/${bucket.key}"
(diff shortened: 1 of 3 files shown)


src/collection-facets.ts
```diff
@@ -352,11 +352,17 @@ export class CollectionFacets extends LitElement {
               facetGroup.key !== 'collection'
                 ? html`${bucket.displayText ?? bucket.key}`
                 : html`
-                    <async-collection-name
-                      .collectionNameCache=${this.collectionNameCache}
-                      .identifier=${bucket.key}
-                      placeholder="-"
-                    ></async-collection-name>
+                    <a
+                      href="/details/${bucket.key}"
```
iisa, line 356: we can hold here for now... when @nsharma123 gets his more facets work in, we can get `baseHost` to prepend

iisa (approved):
LGTM

PR #109, opened 2022-09-13
Morefacets pager updates

iisa:
when starting at 1, show 1 2 3 ... N
image

when under 7 pages, and selected on either [3, 4, 5, 6] => show all pages 1 2 3 4 5 6 7
image

at end, show prev 2
image

shows all pages if count is under 5
image


nsharma123:
working as described

PR #114, opened 2022-09-15
Loading tile styling: adjusted color, gradient direction, border radius

pezvi:


pezvi:
Side-by-side with @iisa — QA check; code good to go

PR #115, opened 2022-09-20
WEBDEV-5470 Migrate to search service v0.4.x (PPS integration)

latonv:
Upgrades the search-service and collection-name-cache dependencies to the newer versions that rely on the Page Production Service.

These changes should not be deployed to the live site until PPS is ready for public beta. Will be kept on alpha tags until everything is ready.


src/collection-browser.ts
```diff
@@ -909,7 +920,7 @@ export class CollectionBrowser
   private async fetchFacets() {
     if (!this.fullQuery) return;
 
-    const aggregations = {
+    /* const aggregations = {
```
dualcnhq, line 923: q: why we removed this part?

also instead of commenting it out, maybe we can totally remove this part of the code or atleast add a note why it is still there.

src/collection-browser.ts
```diff
@@ -936,16 +947,16 @@ export class CollectionBrowser
...
       ],
-    };
+    }; */
 
     const params: SearchParams = {
       query: this.fullQuery,
-      fields: ['identifier'],
-      aggregations,
-      rows: 1,
+      // aggregations, // get default aggregations
```
dualcnhq, line 954: this one as well, better to add note here as well for more context

src/collection-browser.ts
latonv: Ah, good catch -- I'll remove the old comments here and below.

The context for this removal is that the PPS defaults these aggregation parameters based on the page type being requested (here `search_results` is the default page type used). So we don't need to explicitly request them.

latonv: In contrast, the other search call that fetches the search results themselves needs to explicitly omit the aggregations because of this defaulting.

src/collection-browser.ts
latonv: 👍 Contextual note added in latest commit

src/collection-facets/more-facets-content.ts
```diff
@@ -10,23 +10,25 @@ import {
...
+import {
   Aggregation,
   Bucket,
   SearchServiceInterface,
   SearchParams,
+  SearchType,
 } from '@internetarchive/search-service';
 import type { CollectionNameCacheInterface } from '@internetarchive/collection-name-cache';
 import type { ModalManagerInterface } from '@internetarchive/modal-manager';
+import { AggregationSortType } from '@internetarchive/search-service/dist/src/models/aggregation';
```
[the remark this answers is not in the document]
latonv: Thanks for catching that. Looks like SearchService wasn't exporting this type at the top level (but it should be -- will fix).

iisa (approved):
looks good!
small UI pieces I noted during qa:
- desktop sticky side bar shortens when demo is in FTS mode
- More Facets `More...` link should not show in FTS

once those 2 are fixed, we are good to merge! 🎉  :lgtm:

PR #119, opened 2022-09-27
WEBDEV-5487 Fix errors reading facet state from URL params

latonv:
When navigating to a URL that loads collection browser with `and[]` parameters, in certain cases it would throw an error from trying to set the state of facet buckets that didn't exist yet. This fix prevents those errors by constructing temporary facet buckets for the `and[]` parameters with the appropriate state, which will eventually be replaced by the real facet buckets (with correct count) once the search result aggregations load in.

Also adds a bunch of unit tests for the restoration state handler building its state based on the page URL, to prevent future regressions of this nature.

### DIFF
--- a/test/restoration-state-handler.test.ts
+++ b/test/restoration-state-handler.test.ts
@@ -1,0 +1,179 @@
+/* eslint-disable import/no-duplicates */
+import { expect } from '@open-wc/testing';
+import { RestorationStateHandler } from '../src/restoration-state-handler';
+
+import '../src/tiles/grid/tile-stats';
+
+describe('Restoration state handler', () => {
+  it('should restore query from URL', async () => {
+    const handler = new RestorationStateHandler({ context: 'search' });
+
+    const url = new URL(window.location.href);
+    url.search = '?query=boop';
+    window.history.replaceState({ path: url.href }, '', url.href);
+
+    const restorationState = handler.getRestorationState();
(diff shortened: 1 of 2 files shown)


test/restoration-state-handler.test.ts
```diff
@@ -0,0 +1,176 @@
...
+    url.search = '?not[]=year:2018';
+    window.history.replaceState({ path: url.href }, '', url.href);
+
+    const restorationState = handler.getRestorationState();
+    expect(restorationState.selectedFacets.year['2018'].state).to.equal(
+      'hidden'
+    );
+  });
+
+  it('should restore multiple selected/negative facets from URL', async () => {
```
iisa, line 103: 💥

test/restoration-state-handler.test.ts
```diff
@@ -0,0 +1,176 @@
...
+
+    const url = new URL(window.location.href);
+    url.search = '?sort=date';
+    window.history.replaceState({ path: url.href }, '', url.href);
+
+    const restorationState = handler.getRestorationState();
+    expect(restorationState.selectedSort).to.equal('date');
+    expect(restorationState.sortDirection).to.equal('asc');
+  });
+});
```
iisa, line 176: 👏 thank you for bringing this test into the fold :flexalgo:

iisa (approved):
thank you for catching and stabilizing @latonv 🙌 :lgtm: :shipit:

PR #123, opened 2022-09-30
more facets tweak: 35 results/page, 1 col @560px, 3 col after WEBDEV-5489

iisa:
### More modal facets updates:
- Each page displays 35 facets
- Desktop view for the modal:
  - UI: Facets displays in 3 columns, modal has fixed height
  - during paging, there is less vertical scrolling to view the facets on-page
- Mobile view
  - UI: Facets displays in 1 scrollable column per page, modal has fixed height
  - shrink width of modal to see facets displayed in 1 column.
- Scrollable facet group area:
  - when the facet group area is scrollable, its scroll bar does not overlap on facet data
- Note: Modal has max height, will not shrink if window height reduces

#### Testing:
- https://internetarchive.github.io/iaux-collection-browser/pr/pr-123/?query=collection%3Ainlibrary
  - Check "Show Facet Group Outlines" checkbox - each facet group has red top border, blue bottom border
  - Open "More..." **Collection** Facets
  - ✅  when facets wrap, the whole facet block stays together
    - Collection Facets, Page 1, 750px BrowserWidth:  "Washington Research Library Consortium" facet block stays together and wraps to 3rd column (see pic)
image

#### Original Problem:
- [solved]: more facets modal facet group elements does not wrap to next column properly.
  
##### Adjacent issues:
- More facets pager system is currently based on setting N facets per page.
  - Fix: showing 35 results per page reduces the need of "vertical scroll" overflow when there are too many facets displayed at once inside of small modal window.

##### Constraints:
- new paginator behaviour on the way. this update is specifically to up-level current UI.
- when facets are in the modal, they are at a light dom level in <modal-manager>, therefore cannot use coll-brwsr state around mobileView.
  - using CSS media query helps navigate this lack of connection without having to heavily inject the sharedResizeObserver.
  - using "global-level" media query is OK at modal level since there is only 1 modal on the page showing at a time. 👍

### DIFF
--- a/src/collection-facets/facets-template.ts
+++ b/src/collection-facets/facets-template.ts
@@ -232,6 +232,7 @@ export class FacetsTemplate extends LitElement {
         height: auto;
         border-top: var(--facet-row-border-top, 1px solid transparent);
         border-bottom: var(--facet-row-border-bottom, 1px solid transparent);
+        overflow: hidden;
       }
       .facet-info-display {
         display: flex;
--- a/src/collection-facets/facets-template.ts
+++ b/src/collection-facets/facets-template.ts
@@ -190,13 +190,14 @@ export class FacetsTemplate extends LitElement {
 
   static get styles(): CSSResultGroup {
     return css`
+      @media (max-width: 560px) {
+        .facets-on-modal {
+          column-count: 1 !important;
+        }
+      }
       .facets-on-modal {
-        /* For Chrome, Safari, Opera browsers */
-        -webkit-column-width: 100px;
-        /* For Firefox browser */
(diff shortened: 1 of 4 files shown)


iisa:
relates to #125 
will prefer this approach as it also stabilizes desktop facets modal height

bfalling (approved):
LGTM. GitHub notes that "parameters with default values should be last". Should this be addressed?

iisa:
> LGTM. GitHub notes that "parameters with default values should be last". Should this be addressed?

thank u for raising. I've fixed this in another PR #126  - and will merge with this one.

PR #124, opened 2022-09-30
WEBDEV-5488 Lending facets

latonv:
Adds lending ("Availability") facets to the sidebar for parity with legacy search.

Only three facet buckets are shown for lending:
- Lending Library (`is_lendable`)
- Borrow 14 Days (`available_to_borrow`)
- Always Available (`is_readable`)

While there are other possible lending statuses, to my knowledge these are the only three that should be displayed here.


dualcnhq (approved):
LGTM... would be nice if we can address the codecov warnings and GH actions build warnings

src/collection-facets.ts
```diff
@@ -379,6 +402,11 @@ export class CollectionFacets extends LitElement {
       return nothing;
     }
 
+    // Don't render More... links for lending facets
+    if (facetGroup.key === 'lending') {
+      return nothing;
+    }
+
```
iisa, line 409: suggestion: address codecov checks in this file

iisa (approved):
hi! thanks for getting this in quickly 👍 
approach looks good, will look for a few more unit tests to address codecov warnings ^_^
:lgtm:

PR #129, opened 2022-10-04
WEBDEV-5501 Fix bugs on toggling search options

latonv:
- Switching to full text search option shouldn't immediately clear More... facet links (it should wait until new search performed)
- Switching to a different search type should allow performing the same search query again with the new search type. Shouldn't have to change the search query.

### DIFF
--- a/src/collection-browser.ts
+++ b/src/collection-browser.ts
@@ -1183,6 +1183,10 @@ export class CollectionBrowser
    * page are visible, but if the page is not currenlty visible, we don't need to reload
    */
   private get currentVisiblePageNumbers(): number[] {
+    if (!this.infiniteScroller) {
+      return [];
+    }
+
     const visibleCells = this.infiniteScroller.getVisibleCellIndices();
     const visiblePages = new Set<number>();
     visibleCells.forEach(cellIndex => {
--- a/src/collection-browser.ts
+++ b/src/collection-browser.ts
@@ -993,7 +993,7 @@ export class CollectionBrowser
    * If this doesn't change, we don't need to re-fetch the histogram date range
    */
   private get fullQueryNoDateKey() {
-    return `${this.fullQueryWithoutDate}-${this.sortParam?.field}-${this.sortParam?.direction}`;
+    return `${this.fullQueryWithoutDate}-${this.searchType}-${this.sortParam?.field}-${this.sortParam?.direction}`;
   }
 
   /**
--- a/src/collection-browser.ts
+++ b/src/collection-browser.ts
@@ -1183,11 +1183,7 @@ export class CollectionBrowser
    * page are visible, but if the page is not currenlty visible, we don't need to reload
    */
   private get currentVisiblePageNumbers(): number[] {
-    if (!this.infiniteScroller) {
-      return [];
-    }
-
-    const visibleCells = this.infiniteScroller.getVisibleCellIndices();
+    const visibleCells = this.infiniteScroller?.getVisibleCellIndices() ?? [];
     const visiblePages = new Set<number>();
     visibleCells.forEach(cellIndex => {
       const visiblePage = Math.floor(cellIndex / this.pageSize) + 1;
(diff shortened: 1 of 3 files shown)


src/collection-facets.ts
```diff
@@ -398,7 +400,7 @@ export class CollectionFacets extends LitElement {
     facetGroup: FacetGroup
   ): TemplateResult | typeof nothing {
     // Don't render More... links for FTS searches
-    if (this.searchType === SearchType.FULLTEXT) {
+    if (this.previousSearchType === SearchType.FULLTEXT) {
```
iisa, line 403: 👍

src/collection-browser.ts
```diff
@@ -523,6 +538,7 @@ export class CollectionBrowser
         .searchType=${this.searchType}
         .aggregations=${this.aggregations}
         .fullYearsHistogramAggregation=${this.fullYearsHistogramAggregation}
+        .previousSearchType=${this.previousSearchType}
```
iisa, line 541: q: can we reduce the context of `<collection-facets>` by telling it to show its More... links?

src/collection-browser.ts
```diff
@@ -984,7 +1001,7 @@ export class CollectionBrowser
    * If this doesn't change, we don't need to re-fetch the histogram date range
    */
   private get fullQueryNoDateKey() {
-    return `${this.fullQueryWithoutDate}-${this.sortParam?.field}-${this.sortParam?.direction}`;
+    return `${this.fullQueryWithoutDate}-${this.searchType}-${this.sortParam?.field}-${this.sortParam?.direction}`;
```
iisa, line 1004: Q: what does this update do?

src/collection-browser.ts
```diff
@@ -271,6 +278,14 @@ export class CollectionBrowser
     this.sortDirection = null;
   }
 
+  /**
+   * Manually requests to perform a search, which will only be executed if one of
+   * the query, the search type, or the sort has changed.
+   */
+  requestSearch() {
```
iisa, line 285: Q: where does this get called?

src/collection-browser.ts
latonv: From what I understand, this key is used to determine whether the year histogram needs to be re-fetched. So if someone selects a date facet, the year histogram can remain unchanged because date facets don't affect this key.

However, a different search type *does* affect the need to re-fetch the year histogram (along with all the other results/facets), so it should be part of this key.

src/collection-browser.ts
latonv: This is intended to be called by Offshoot code, so nowhere in collection-browser itself.

The use case is that someone has selected a different search type (say, switching from metadata to full text). Selecting that different option doesn't trigger a search on its own, but we still need a way to request that search later on when they hit Go, even if the search query itself hasn't changed.

iisa: q: is there an interface that can represent this external api relationship by chance?

src/collection-browser.ts
```diff
@@ -523,6 +538,7 @@ export class CollectionBrowser
         .searchType=${this.searchType}
         .aggregations=${this.aggregations}
         .fullYearsHistogramAggregation=${this.fullYearsHistogramAggregation}
+        .moreLinksVisible=${this.previousSearchType !== SearchType.FULLTEXT}
```
iisa, line 541: 👍

src/collection-browser.ts
iisa: ah is this the `sin=` support?

iisa (approved):
looks good thank you for updates. 
just got some clarifying questions there, everything looks good to merge, thank u.

src/collection-browser.ts
latonv: Not related to the `sin` param support, no. This (and the related change on line 1066) is essentially a UX fix to allow patrons to change the search type and perform a new search without changing the search query. Without this change, collection browser doesn't update the search results if someone just changes the search type (because it relies on these keys to avoid duplicate searches).

src/collection-browser.ts
latonv: I think there's a larger question here around more clearly specifying CollectionBrowser's external interface. At least from Offshoot's perspective, that interface also includes the `clearFilters` method and a number of other properties. We should definitely think about what we want this to look like going forward, but I'd rather limit this PR's scope and address these questions of architecture in the next one.

PR #131, opened 2022-10-05
Fix loggedIn icon overlay in grid and compact display mode

dualcnhq:
When logged out, the message should be "Log in to view this content" and the icon should be the login icon.

When logged in, the message should say "Content may be inappropriate" and the icon should be the caution triangle


latonv (approved):
LGTM! 👍

PR #132, opened 2022-10-05
WEBDEV-5509 Add `sin` URL param to restoration state

latonv:
This PR makes the restoration state handler load and save the `sin` param that determines the search type of a results page.

The demo app is also updated to make the dev tools panel more compact and readable, and to handle some other URL params that will allow easier testing of the search service through collection browser demos.

### DIFF
--- a/src/restoration-state-handler.ts
+++ b/src/restoration-state-handler.ts
@@ -189,7 +189,7 @@ export class RestorationStateHandler
 
     if (searchInside) {
       restorationState.searchType =
-        searchInside === 'TXT' ? SearchType.FULLTEXT : SearchType.METADATA;
+        searchInside === 'TXT' ? SearchType.FULLTEXT : undefined; // No explicit metadata sin
     }
     if (pageNumber) {
       const parsed = parseInt(pageNumber, 10);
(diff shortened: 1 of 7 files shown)


src/restoration-state-handler.ts
```diff
@@ -173,6 +187,10 @@ export class RestorationStateHandler
       },
     };
 
+    if (searchInside) {
+      restorationState.searchType =
+        searchInside === 'TXT' ? SearchType.FULLTEXT : undefined; // No explicit metadata sin
```
iisa, line 192: Q: how can we get it so that when undefined, the `sin=TXT` gets removed from url?

iisa (approved):
looks good!
just had a Question about demo behavior that we want to match on offshoot app - `sin=TXT` did not get updated in Url when toggling to & running MD search.. once that is good, then we may have pattern/structure to take to offshoot with us.

Q: you've got a lot of great stuff in play, what's the best way to help test on an Offshoot review app?

src/restoration-state-handler.ts
latonv: Should be fixed in latest commit

PR #134, opened 2022-10-07
WEBDEV-5520 Reinstate previously-missing search result fields

latonv:
A handful of search result fields that we used to fetch from `advancedsearch.php` were initially not available on PPS result hits. However, these are now being served by the PPS and can be used to populate the tile models again.
- `addeddate`
- `avg_rating`
- `issue`
- `item_count`
- `source`

This PR reinstates support for these fields so that they appear correctly on result tiles/list items.


iisa (approved):
:lgtm:

PR #135, opened 2022-10-08
WEBDEV-5513 Don't render certain collections as facets or list items

latonv:
Certain meta-collections (currently `deemphasize` and `community`) shouldn't be rendered in the facet sidebar or in item collection lists. This PR ensures such collections are not displayed in those two places.

### DIFF
--- a/src/collection-browser.ts
+++ b/src/collection-browser.ts
@@ -946,6 +946,8 @@ export class CollectionBrowser
     const params: SearchParams = {
       query: this.fullQuery,
       rows: 0,
+      // Fetch a few extra buckets beyond the 6 we show, in case some get suppressed
+      aggregationsSize: 10,
       // Note: we don't need an aggregations param to fetch the default aggregations from the PPS.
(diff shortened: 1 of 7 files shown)


src/collection-browser.ts
```diff
@@ -1088,27 +1088,6 @@ export class CollectionBrowser
     const sortParams = this.sortParam ? [this.sortParam] : [];
     const params: SearchParams = {
       query: this.fullQuery,
-      fields: [
```
iisa, line 1091: Q: where did this go?

iisa:
merge ok pending QA questions/concerns. more details in jira ticket

src/collection-browser.ts
latonv: The fields param is no longer necessary since the PPS responses have all of these fields by default. But I think this change has since become part of another PR, so I'll undo the change here.

iisa (approved):
👍

PR #138, opened 2022-10-11
WEBDEV-5518 Change all links color

dualcnhq:
### Testing Steps

- go and search for `puppies`
- switch to list view
- note the link color is incorrect for links to “More…” facets, item titles, creators, topics, collections

   - Color is #0000ee, should be #4b64ff

image

**Updated:**
Screen Shot 2022-10-12 at 04 15 17

### DIFF
--- a/src/collection-facets.ts
+++ b/src/collection-facets.ts
@@ -585,7 +585,7 @@ export class CollectionFacets extends LitElement {
         padding: 0;
         background: inherit;
         border: 0;
-        color: #4b64ff;
+        color: var(--ia-theme-link-color, #4b64ff);
         cursor: pointer;
       }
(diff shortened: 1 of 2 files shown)


src/collection-facets.ts
```diff
@@ -585,9 +585,10 @@ export class CollectionFacets extends LitElement {
         padding: 0;
         background: inherit;
         border: 0;
-        color: blue;
+        color: #4b64ff;
```
iisa, line 588: can we get a var? `--collectionBrowserLinkColor` or something so that all the other anchor colored links get the same treatment?

iisa (approved):
:lgtm:

PR #139, opened 2022-10-11
WEBDEV-5512 Fix alignment of image and title text in listview mode

dualcnhq:
### Testing steps:

- go search for `happy face`
- switch to list view display mode 

Note - this issue usually occurs when:
- there’s no default thumbnail/image is set in an item or it has the default IA logo with black background
- the thumbnail image is not 100px height size

**Previous:**

image

**New:**
Screen Shot 2022-10-12 at 04 02 54


iisa:
looks good! Q: how do we make the change so that test coverage doesn't drop?

PR #140, opened 2022-10-12
WEBDEV-5499: Merge all facet requests into a single request for the default list

latonv:
The PPS now includes lending and year_histogram aggregations in the default list of aggregations, so these no longer need to be fetched separately. This PR makes use of these bundled aggregations and removes the old additional requests.

### DIFF
--- a/test/collection-browser.test.ts
+++ b/test/collection-browser.test.ts
@@ -154,8 +154,11 @@ describe('Collection Browser', () => {
   });
 
   it('should render with a sort bar, facets, and infinite scroller', async () => {
+    const searchService = new MockSearchService();
+
     const el = await fixture<CollectionBrowser>(
-      html`<collection-browser></collection-browser>`
+      html`<collection-browser .searchService=${searchService}>
+      </collection-browser>`
     );
 
     el.baseQuery = 'hello';
@@ -238,6 +241,9 @@ describe('Collection Browser', () => {
 
     // This shouldn't throw an error
     expect(el.fetchPage(2)).to.exist;
+
+    // Should continue showing the empty placeholder
+    expect(el.shadowRoot?.querySelector('empty-placeholder')).to.exist;
   });
 
   it('applies loggedin flag to tile models if needed', async () => {
@@ -444,8 +450,11 @@ describe('Collection Browser', () => {
   });
 
   it('sets sort properties when user changes sort', async () => {
+    const searchService = new MockSearchService();
+
     const el = await fixture<CollectionBrowser>(
-      html`<collection-browser></collection-browser>`
+      html`<collection-browser .searchService=${searchService}>
+      </collection-browser>`
     );
 
     expect(el.selectedSort).to.equal(SortField.relevance);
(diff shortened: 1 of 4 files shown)


test/collection-browser.test.ts
```diff
@@ -227,6 +228,79 @@ describe('Collection Browser', () => {
...
+  it('fails gracefully if no search service provided', async () => {
+    const el = await fixture<CollectionBrowser>(
+      html`<collection-browser></collection-browser>`
+    );
+
+    el.baseQuery = 'collection:foo';
+    await el.updateComplete;
+
+    // This shouldn't throw an error
+    expect(el.fetchPage(2)).to.exist;
```
iisa, line 240: Q: what does UI do?

test/collection-browser.test.ts
```diff
@@ -227,6 +228,79 @@ describe('Collection Browser', () => {
...
+  });
+
+  it('applies loggedin flag to tile models if needed', async () => {
```
iisa, line 243: 👍

test/collection-browser.test.ts
```diff
@@ -227,6 +228,79 @@ describe('Collection Browser', () => {
...
+    const cellTemplate = el.cellForIndex(0);
+    expect(cellTemplate).to.exist;
+
+    const cell = await fixture<TileDispatcher>(cellTemplate!);
+    expect(cell).to.exist;
+
+    expect(cell.model?.contentWarning).to.be.true;
+  });
+
+  it('both loggedin and no-preview flags can be set simultaneously', async () => {
```
iisa, line 283: 👍

test/collection-browser.test.ts
```diff
@@ -227,6 +228,79 @@ describe('Collection Browser', () => {
...
+    await el.updateComplete;
+
+    const cellTemplate = el.cellForIndex(0);
+    expect(cellTemplate).to.exist;
+
+    const cell = await fixture<TileDispatcher>(cellTemplate!);
+    expect(cell).to.exist;
+
+    expect(cell.model?.loginRequired).to.be.true;
+    expect(cell.model?.contentWarning).to.be.true;
```
iisa, line 301: we check the cell's tile model...what else other ways can we validate the test statement, "both loggedin and no-preview flags can be set simultaneously?"

do we check for the UI piece?

iisa (approved):
looks good! always nice to prune files.

thanks for the start on unit tests.  There may be even more concrete assertions we can make given the new test statements aside from confirming model references. not a blocker, just a note and light challenge on creating more expressive assertions...

test/collection-browser.test.ts
latonv: Yeah, these could be doing more with the UI for sure. I think my high-level goal with here was to ensure we're testing for correct functionality in collection-browser's `updateDataSource` method that sets these flags from the search response, rather than testing the actual UI result (which should probably happen in the tile tests, not collection-browser itself).

test/collection-browser.test.ts
latonv: I believe it should just leave the empty placeholder up because the data source never gets updated if no search occurs. But I'll update the test to verify that!

PR #143, opened 2022-10-13
WEBDEV-5532 Show weekly view counts on list items when sorting by them

latonv:
Items in both list view and compact view are incorrectly showing all-time view counts even when sorting by weekly views. This PR adds weekly views to the tile models and displays them on list items when sorting by weekly views.


dualcnhq (approved):
LGTM

PR #150, opened 2022-10-15
WEBDEV-5546 Only show placeholder for date picker when it's enabled

latonv:
With the facet tombstones update, there’s a placeholder block being shown for the date picker while it loads. However, when the date picker is disabled, that placeholder is confusing and shouldn’t show up. This PR ensures the facet tombstones do not include a placeholder block for the date picker when the feature is disabled.


dualcnhq (approved):

iisa (approved):
:shipit:

PR #154, opened 2022-10-26
WEBDEV-5567 Show 3-line descriptions

latonv:
These changes ensure that the tile model receives the full description value (which may be an array consisting of many lines), rather than just the first line. Then, when the description is rendered in extended list view, this PR ensures these line breaks are correctly rendered and each line is correctly sanitized. While only 3 lines of description will ever be visible in the tile preview, this ensures we are not prematurely truncating the preview at the first line break.

This PR also updates the demo app & search service version to avoid running into CORS errors when specifying a custom `search_base_url` (e.g., for testing against a review app).

### DIFF
--- a/package.json
+++ b/package.json
@@ -23,7 +23,7 @@
   "types": "dist/index.d.ts",
   "dependencies": {
     "@internetarchive/analytics-manager": "^0.1.2",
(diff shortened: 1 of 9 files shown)


iisa (approved):
looks good

PR #155, opened 2022-10-27
WEBDEV-5549 Show placeholder tiles upon each new search performed

latonv:
When going from a blank search (empty placeholder) to a new search, the “loading placeholder” tile tombstones appear while the results load, as expected. But if you perform another search after that, the tile tombstones don’t appear anymore – the old results stay up until the new ones arrive. So there’s no indication that results are loading.

This PR ensures the search results always show placeholder tombstones immediately upon performing a new search, regardless of whether or not there were other search results present beforehand.


bfalling (approved):
LGTM

PR #156, opened 2022-10-28
WEBDEV-5572 Only send separate year_histogram requests when necessary

latonv:
Currently a separate PPS request is sent to obtain year histogram data for the date picker, regardless of whether the date picker is present on the page.

Moreover, the request that retrieves all the other facets also includes the year histogram data, which is correct as long as there are no year facets or date range filters applied. In those cases, there is no need for a separate year histogram request. The additional request is only necessary if the user has selected a year facet or specified a date range.

This PR ensures a separate year histogram facet request is only sent if the date picker is actually enabled and there is a selected year facet or a date range filter applied. It also refactors some of the facet-query-building getters to decouple the with-date query from the without-date one since these logic fixes make the existing "build one from the other" approach a bit messy.

### DIFF
--- a/src/app-root.ts
+++ b/src/app-root.ts
@@ -421,6 +421,13 @@ export class AppRoot extends LitElement {
   private datePickerChanged(e: Event) {
     const target = e.target as HTMLInputElement;
     this.collectionBrowser.showHistogramDatePicker = target.checked;
+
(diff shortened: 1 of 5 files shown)


bfalling (approved):
LGTM

PR #157, opened 2022-10-31
WEBDEV-5575 Clarify loading state and empty search data model on new search

latonv:
This PR changes the "- Results" label to reflect the loading state by changing to "Searching..." when a new set of results is loading as a result of a query change. The data model (incl. results, total result count, and aggregations) is now also fully emptied as soon as a new search is performed, rather than persisting until the new search results arrive.

### DIFF
--- a/test/collection-browser.test.ts
+++ b/test/collection-browser.test.ts
@@ -20,7 +20,15 @@ import { analyticsCategories } from '../src/utils/analytics-events';
 import type { TileDispatcher } from '../src/tiles/tile-dispatcher';
 import type { CollectionFacets } from '../src/collection-facets';
 
-/** Wait for the next tick of the event loop */
+/**
+ * Wait for the next tick of the event loop.
+ *
+ * This is necessary in some of the tests because certain collection browser
+ * updates take more than one tick to render (e.g., date picker & query changes).
+ * These delays are non-ideal and should eventually be investigated and fixed,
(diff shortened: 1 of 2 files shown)


test/collection-browser.test.ts
```diff
@@ -20,6 +20,12 @@ import { analyticsCategories } from '../src/utils/analytics-events';
 import type { TileDispatcher } from '../src/tiles/tile-dispatcher';
 import type { CollectionFacets } from '../src/collection-facets';
 
+/** Wait for the next tick of the event loop */
```
mc2, line 23: Question: This looks interesting and useful, but why is it being used and useful?
Suggest additional "why" comment.

mc2:
https://webarchive.jira.com/browse/WEBDEV-5575

test/collection-browser.test.ts
latonv: With collection browser being a large and complex beast, I've noticed there are some parts of it that seem to take multiple component update cycles to trickle through before a change is rendered. One example of this is the date picker, which seems to have some unusual logic that delays it from rendering in the same tick as certain updates. So in cases like that, we just need to wait an additional tick before we can reliably expect our test conditions to have rendered completely.

I don't think that's an ideal situation -- it's probably a good target for a refactor at some point. But anyway, I can add some extra comments clarifying this!

PR #159, opened 2022-11-02
WEBDEV-5581 Hide fav- collections in list view tiles

latonv:
The list of collection links on tiles in extended list view should not include "user favorites" collections (`fav-<username>`). This PR ensures those collections are not rendered.


nsharma123 (approved):
LGTM!

PR #163, opened 2022-11-10
WEBDEV-5599 Prevent item images from having undefined URLs

latonv:
Currently, if search result tiles are rendered before the page config has fully loaded, they may have an invalid `baseImageUrl` which they use to fetch their item image, resulting in many 404s for URLs containing `undefined`.

This PR adds a guard clause preventing these bad image requests from firing before the config (with correct `baseImageUrl`) has properly loaded.


iisa (approved):
:lgtm:

PR #167, opened 2022-11-16
WEBDEV-5561 Disable/gray-out letters with 0 results in alphabet filter bar

latonv:
On the legacy search page, when the alphabet filter bar is shown, letters which correspond to 0 items are grayed out and unclickable. This PR adds that functionality to the collection browser.

### DIFF
--- a/src/collection-browser.ts
+++ b/src/collection-browser.ts
@@ -664,7 +664,7 @@ export class CollectionBrowser
       changed.has('dateRangeQueryClause') ||
       changed.has('selectedFacets')
     ) {
-      this.clearLetterCounts();
+      this.refreshLetterCounts();
     }
     if (changed.has('selectedSort') || changed.has('sortDirection')) {
       const prevSortDirection = changed.get('sortDirection') as SortDirection;
@@ -1402,12 +1402,21 @@ export class CollectionBrowser
   }
 
   /**
-   * Sets both the title and creator letter counts to undefined.
-   * Call this whenever they are invalidated (e.g., the query changed).
+   * Clears the cached letter counts for both title and creator, and
+   * fetches a new set of counts for whichever of them is the currently
+   * selected sort option (which may be neither).
+   *
+   * Call this whenever the counts are invalidated (e.g., by a query change).
    */
-  private clearLetterCounts(): void {
+  private refreshLetterCounts(): void {
     this.titleLetterCounts = undefined;
     this.creatorLetterCounts = undefined;
+
+    if (this.selectedSort === 'title') {
+      this.fetchTitleLetterCounts();
+    } else if (this.selectedSort === 'creator') {
+      this.fetchCreatorLetterCounts();
+    }
   }
 
   /*
--- a/src/collection-browser.ts
+++ b/src/collection-browser.ts
@@ -48,6 +48,9 @@ import {
   CollectionDisplayMode,
   FacetOption,
   FacetBucket,
+  PrefixFilterType,
+  PrefixFilterCounts,
+  prefixFilterAggregationKeys,
 } from './models';
 import {
   RestorationStateHandlerInterface,
@@ -178,9 +181,10 @@ export class CollectionBrowser
 
   @state() private placeholderType: PlaceholderType = null;
 
-  @state() private titleLetterCounts?: Record<string, number>;
-
-  @state() private creatorLetterCounts?: Record<string, number>;
+  @state() private prefixFilterCountMap: Record<
+    PrefixFilterType,
+    PrefixFilterCounts
+  > = { title: {}, creator: {} };
 
   @query('#content-container') private contentContainer!: HTMLDivElement;
 
@@ -381,8 +385,7 @@ export class CollectionBrowser
         .displayMode=${this.displayMode}
         .selectedTitleFilter=${this.selectedTitleFilter}
         .selectedCreatorFilter=${this.selectedCreatorFilter}
-        .titleLetterCounts=${this.titleLetterCounts}
-        .creatorLetterCounts=${this.creatorLetterCounts}
+        .prefixFilterCountMap=${this.prefixFilterCountMap}
         .resizeObserver=${this.resizeObserver}
         @sortChanged=${this.userChangedSort}
         @displayModeChanged=${this.displayModeChanged}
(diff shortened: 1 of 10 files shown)


src/collection-browser.ts
```diff
@@ -423,6 +430,13 @@ export class CollectionBrowser
 
     if (!sortField) return;
     this.sortParam = { field: sortField, direction: this.sortDirection };
+
+    // Lazy-load the alphabet counts for title/creator sort bar as needed
+    if (this.selectedSort === 'title' && !this.titleLetterCounts) {
```
mc2, line 435: Question: for DRY, `fetchTitleLetterCounts()` and `fetchCreatorLetterCounts()` seem remarkably similar. With above line 384 comment, could we simply have 
```
fetchFirstCharCounts();
```
 here? And handle all the logic inside that function, as needed by `this.selectedSort` and the other var(s), including doing nothing?

src/collection-browser.ts
```diff
@@ -376,6 +381,8 @@ export class CollectionBrowser
         .displayMode=${this.displayMode}
         .selectedTitleFilter=${this.selectedTitleFilter}
         .selectedCreatorFilter=${this.selectedCreatorFilter}
+        .titleLetterCounts=${this.titleLetterCounts}
```
mc2, line 384: Question: Do we really need counts for both title and creator? Would a single `.charCounts` var work?

src/collection-browser.ts
```diff
@@ -1320,6 +1357,71 @@ export class CollectionBrowser
...
+  }
+
+  /**
+   * Clears the cached letter counts for both title and creator, and
+   * fetches a new set of counts for whichever of them is the currently
+   * selected sort option (which may be neither).
+   *
+   * Call this whenever the counts are invalidated (e.g., by a query change).
+   */
+  private refreshLetterCounts(): void {
```
mc2, line 1414: Question: As above, do we really need both vars and both methods?

src/sort-filter-bar/sort-filter-bar.ts
```diff
@@ -42,6 +42,10 @@ export class SortFilterBar
 
   @property({ type: Boolean }) showRelevance: boolean = true;
 
+  @property({ type: Object }) titleLetterCounts?: Record<string, number>;
```
mc2, line 45: Question: As above, do we really need both vars and both methods?

src/sort-filter-bar/sort-filter-bar.ts
```diff
@@ -563,13 +567,15 @@ export class SortFilterBar
   private get titleSelectorBar() {
     return html` <alpha-bar
       .selectedLetter=${this.selectedTitleFilter}
+      .letterCounts=${this.titleLetterCounts}
```
mc2, line 570: Question: As above, do we really need both vars and both methods?

test/mocks/mock-search-service.ts
```diff
@@ -15,6 +15,8 @@ import {
   mockSuccessLoggedInAndNoPreviewResult,
   mockSuccessWithYearHistogramAggs,
   mockSuccessMultiLineDescription,
+  mockSuccessFirstTitleResult,
```
mc2, line 18: Question: As above, do we really need both vars and both methods?

test/sort-filter-bar/sort-filter-bar.test.ts
```diff
@@ -185,3 +185,47 @@ describe('Display mode/style buttons', () => {
     expect(displayModeTitle).to.equal('Tile view');
   });
 });
+
+describe('Sort/filter bar letter behavior', () => {
+  it('sets the selected title letter when clicked', async () => {
```
mc2, line 190: Question: As above, do we really need both vars and both methods?

src/collection-browser.ts
mc2, line 384: Note: reviewing to the end, it seems like the Title letter picker UI and Creator letter picker UI already use separate `titleLetterCounts` and `creatorLetterCounts` -- but I haven't looked at the UI. 

Maybe combining these has implications for existing UI code -- and my comments are, if not misplaced, than are not helpful because of constraints of existing code.

latonv: My thinking around this was, since generating aggregation data tends to be rather expensive on the backend, we should only (lazily) request what we actually need to avoid unnecessary overhead, and should cache the counts to avoid having to recompute them if users switch between title/creator sorting. Having them separate avoids the need to repeatedly request & overwrite the same data when people change the sort type.

That said, maybe I'm just prematurely optimizing... it's possible the `firstX` counts aren't actually as expensive as other aggregation types.

mc2: >avoid having to recompute them if users switch between title/creator sorting
Thanks @latonv. This makes sense to me.

mc2: Will resolve relevant conversations.

src/collection-browser.ts
```diff
@@ -1320,6 +1357,71 @@ export class CollectionBrowser
...
+  private async fetchLetterBuckets(
+    type: 'firstTitle' | 'firstCreator'
+  ): Promise<Bucket[]> {
+    if (!this.fullQueryWithoutAlphaFilters) return [];
+
+    const params: SearchParams = {
+      query: this.fullQueryWithoutAlphaFilters,
+      rows: 0,
+      // Only fetch the firstTitle or firstCreator aggregation
+      aggregations: { simpleParams: [type] },
```
mc2, line 1369: Question: In reading/review, I was looking for the definition of `simpleParams`. Didn't find it. Low priority.

src/collection-browser.ts
```diff
@@ -1320,6 +1357,71 @@ export class CollectionBrowser
     }
   }
 
+  private async fetchLetterBuckets(
```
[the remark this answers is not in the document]
latonv: Thanks for the suggestion! I've refactored around the notion of "prefix filters" in an attempt to better capture the general case which might eventually include 0-9 or others.

src/collection-browser.ts
```diff
@@ -1320,6 +1357,71 @@ export class CollectionBrowser
...
+    type: 'firstTitle' | 'firstCreator'
```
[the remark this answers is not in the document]
latonv: Something we've done with other facet types is have a single source of truth (in our models) for mapping between the UI concepts and backend keys. So I've updated to use a similar strategy here, which serves to decouple the UI components from these backend implementation details.

mc2 (approved):

PR #168, opened 2022-11-16
WEBDEV-5620 Fix page fetch race condition that sometimes causes tiles not to load

latonv:
The page fetch logic currently assumes that when the last page of results loads in, it must represent the final page fetch. So it caps off the infinite scroller’s tile count with whatever count it currently has.

However, if the last couple of pages load out-of-order (e.g., page N arrives before page N-1), then this prematurely limits the number of items shown (page N-1 never shows up in the UI).

This is particularly problematic on [searches that have just over 50 items](https://internetarchive.github.io/iaux-collection-browser/main/?query=pro+archia+poeta) – if the second page of items arrives before the first, then only a few items will ever be shown, completely ignoring the first 50.

This PR addresses the issue by ensuring that the infinite scroller's item count is not capped off until all pending page fetches resolve.

### DIFF
--- a/src/collection-browser.ts
+++ b/src/collection-browser.ts
@@ -1211,10 +1211,19 @@ export class CollectionBrowser
       this.preloadCollectionNames(results);
       this.updateDataSource(pageNumber, results);
     }
-    if (results.length < this.pageSize) {
+
+    // When we reach the end of the data, we can't always immediately cap off the
+    // tile count, because there may be previous pages that haven't finished loading
+    // yet. So we flag the end of the data, but let the final page load set the count.
+    // This avoids race conditions dependent on the order pages arrive in.
+    if (results.length < this.pageSize || this.endOfDataReached) {
       this.endOfDataReached = true;
-      // this updates the infinite scroller to show the actual size
-      if (this.infiniteScroller) {
+      // This updates the infinite scroller to show the actual size
+      // (provided there are no other pages still being fetched)
+      if (
+        this.infiniteScroller &&
+        Object.keys(this.pageFetchesInProgress).length === 1 // This must be the last page fetch
+      ) {
         this.infiniteScroller.itemCount = this.actualTileCount;
       }
     }
--- a/src/collection-browser.ts
+++ b/src/collection-browser.ts
@@ -1212,20 +1212,12 @@ export class CollectionBrowser
       this.updateDataSource(pageNumber, results);
     }
 
-    // When we reach the end of the data, we can't always immediately cap off the
-    // tile count, because there may be previous pages that haven't finished loading
-    // yet. So we flag the end of the data, but let the final page load set the count.
-    // This avoids race conditions dependent on the order pages arrive in.
-    if (results.length < this.pageSize || this.endOfDataReached) {
+    // When we reach the end of the data, we can set the infinite scroller's
+    // item count to the real total number of results (rather than the
+    // temporary estimates based on pages rendered so far).
+    if (results.length < this.pageSize) {
       this.endOfDataReached = true;
-      // This updates the infinite scroller to show the actual size
-      // (provided there are no other pages still being fetched)
-      if (
-        this.infiniteScroller &&
-        Object.keys(this.pageFetchesInProgress).length === 1 // This must be the last page fetch
-      ) {
-        this.infiniteScroller.itemCount = this.actualTileCount;
-      }
+      this.infiniteScroller.itemCount = this.totalResults;
     }
     this.pageFetchesInProgress[pageFetchQueryKey]?.delete(pageNumber);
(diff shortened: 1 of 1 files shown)


src/collection-browser.ts
```diff
@@ -1211,10 +1211,19 @@ export class CollectionBrowser
       this.preloadCollectionNames(results);
       this.updateDataSource(pageNumber, results);
     }
-    if (results.length < this.pageSize) {
+
+    // When we reach the end of the data, we can't always immediately cap off the
+    // tile count, because there may be previous pages that haven't finished loading
+    // yet. So we flag the end of the data, but let the final page load set the count.
+    // This avoids race conditions dependent on the order pages arrive in.
+    if (results.length < this.pageSize || this.endOfDataReached) {
```
iisa, line 1219: Q: is there a way to track/know how many pages one has scrolled through?

src/collection-browser.ts
```diff
@@ -1211,10 +1211,19 @@ export class CollectionBrowser
...
       this.endOfDataReached = true;
-      // this updates the infinite scroller to show the actual size
-      if (this.infiniteScroller) {
+      // This updates the infinite scroller to show the actual size
+      // (provided there are no other pages still being fetched)
+      if (
+        this.infiniteScroller &&
```
iisa, line 1224: Q: do we the need to check if `this.infiniteScroller`? feels like the `infiniteScroller` should be absolute..

src/collection-browser.ts
```diff
@@ -1211,10 +1211,19 @@ export class CollectionBrowser
...
+        Object.keys(this.pageFetchesInProgress).length === 1 // This must be the last page fetch
```
iisa, line 1225: Q: is `Object.keys(this.pageFetchesInProgress).length === 1` equal to `results.length < this.pageSize || this.endOfDataReached`?

iisa (commented):
heck of a find!
had some questions around conditional values used, moreso for knowledge share.

how hard is this to test?

src/collection-browser.ts
latonv: From what I understand, `this.pagesToRender` is (roughly) a measure of how many pages have been scrolled through. It's set to 1 whenever the query changes, and is incremented each time the scroll threshold for loading a new page is reached.

src/collection-browser.ts
latonv: Good question! The infinite scroller should always exist _except_ when a "no query"/"no results" placeholder is shown. And that placeholder should be cleared before we start fetching a page. So I don't _think_ it's possible to reach this point without the infinite scroller existing.

It's possible this guard was just added out of an abundance of caution, since `this.infiniteScroller` is defined by a `@query` decorator.

src/collection-browser.ts
latonv: The `Object.keys(this.pageFetchesInProgress).length === 1` condition ensures that no other page fetches are pending, because we don't want to prematurely set the infinite scroller's final item count before all remaining fetches have finished.

The earlier `results.length < this.pageSize || this.endOfDataReached` condition is checking whether we've reached the last page of results. `results.length < this.pageSize` will be true if the page we're currently processing is the last one (or beyond), while `this.endOfDataReached` will be true if we've _previously_ processed the last page (e.g., if the page responses have arrived out of order).

latonv: Although now I'm wondering why the infinite scroller's item count was set in such a roundabout way, despite having the total result count present on every response (even going back to the `advancedsearch.php` days...)

latonv:
> heck of a find! had some questions around conditional values used, moreso for knowledge share.
> 
> how hard is this to test?

Thanks @iisa for the great questions! Made me rethink through the logic and see that much of it was obsolete (probably only relevant for `advancedsearch.php` times). So now there's a much simpler fix for the race condition.

I think to test this, we might need to beef up our mock search service with a way to add arbitrary delays to specific fetches... without that, I don't think we currently have a good way to write tests that reliably hit the race condition.

src/collection-browser.ts
```diff
@@ -1211,12 +1211,13 @@ export class CollectionBrowser
...
+    // When we reach the end of the data, we can set the infinite scroller's
+    // item count to the real total number of results (rather than the
+    // temporary estimates based on pages rendered so far).
     if (results.length < this.pageSize) {
       this.endOfDataReached = true;
-      // this updates the infinite scroller to show the actual size
-      if (this.infiniteScroller) {
-        this.infiniteScroller.itemCount = this.actualTileCount;
-      }
+      this.infiniteScroller.itemCount = this.totalResults;
```
iisa, line 1220: 👍

iisa (approved):
clean ✨

PR #172, opened 2022-11-23
New item tile with flexible height

dualcnhq:


src/collection-browser.ts
```diff
@@ -1801,14 +1801,6 @@ export class CollectionBrowser
         18rem
       );
       --infiniteScrollerCellMaxWidth: var(--collectionBrowserCellMaxWidth, 1fr);
-      --infiniteScrollerCellMinHeight: var(
-        --collectionBrowserCellMinHeight,
-        29rem
-      );
-      --infiniteScrollerCellMaxHeight: var(
-        --collectionBrowserCellMaxHeight,
-        29rem
```
iisa, line 1810: great find

src/tiles/grid/item-tile.ts
```diff
@@ -75,10 +76,14 @@ export class ItemTile extends LitElement {
...
-  private get doesSortedByDate() {
-    return ['date', 'reviewdate', 'addeddate', 'publicdate'].includes(
-      this.sortParam?.field as string
-    );
+  private get creatorTemplate(): TemplateResult | typeof nothing {
+    if (!this.model?.creator) return nothing;
+
+    return html`
+      <div class="created-by">
+        <span class="truncated">by&nbsp;${this.model?.creator}</span>
```
iisa, line 84: light suggestion: maybe we add `title` attribute here to distinguish what the data means `title='creator'`?
https://www.w3schools.com/tags/att_global_title.asp

src/tiles/grid/item-tile.ts
```diff
@@ -110,47 +115,63 @@ export class ItemTile extends LitElement {
...
+  private get volumeIssueTemplate(): TemplateResult | typeof nothing {
+    if (!this.model?.volume || !this.model?.issue) return nothing;
 
-    return html`<text-snippet-block
-      viewsize="grid"
-      .snippets=${this.model?.snippets}
-    ></text-snippet-block>`;
+    return html`
+      <div class="volume-issue">
+        <span class="truncated">
```
iisa, line 132: same here with title='volume|foo' suggestion as above

src/tiles/grid/tile-stats.ts
```diff
@@ -93,14 +93,15 @@ export class TileStats extends LitElement {
...
       }
 
       #stats-row {
-        border-top: 1px solid #bbb;
-        align-items: center;
+        align-items: left;
         display: flex;
         flex: 1;
-        justify-content: space-evenly;
+        justify-content: space-between;
```
iisa, line 104: does this give the extra padding that was needed?

src/tiles/image-block.ts
```diff
@@ -86,12 +86,14 @@ export class ImageBlock extends LitElement {
         display: flex;
         justify-content: center;
         position: relative;
+        background-color: #f1f1f4;
```
iisa, line 89: Q: what color is this? is this special to the tile itself (do we use it anywhere else?)?

src/tiles/grid/item-tile.ts
```diff
@@ -35,38 +35,39 @@ export class ItemTile extends LitElement {
...
+              .baseImageUrl=${this.baseImageUrl}
+              .loggedIn=${this.loggedIn}
+              .isCompactTile=${false}
+              .isListTile=${false}
+              .viewSize=${'grid'}
+            >
+            </image-block>
+
+            <div id="title">
+              <h1 class="truncated" title=${ifDefined(itemTitle)}>
```
iisa, line 52: I like how `title=${ifDefined(itemTitle)` is set here 👍

iisa (commented):
looks good, some questions, comments.
will defer to you and @latonv for final push ^_^

src/tiles/grid/tile-stats.ts
```diff
@@ -93,14 +93,15 @@ export class TileStats extends LitElement {
...
       .item-stats {
         height: 35px;
+        padding-left: 5px;
+        padding-right: 5px;
       }
 
       #stats-row {
-        border-top: 1px solid #bbb;
-        align-items: center;
+        align-items: left;
```
latonv, line 101: I don't think `align-items: left` is a valid rule here. Did you mean `text-align: left`?

Is this meant to adjust the mediatype icon positioning? I notice they seem a bit misaligned so maybe this will fix it? 😄

latonv (commented):
Looks good! Noted a couple CSS things that might need minor adjustments.

src/tiles/grid/tile-stats.ts
latonv: Oh, I see now there's already a `text-align: center` rule below, which seems correct. And now I'm finding the mediatype icons are aligned properly in the demo app, so I probably just messed something up while testing. Oops! 🙃 

So this should probably still be `align-items: center` like before, right?

src/tiles/image-block.ts
dualcnhq: just for this one

src/tiles/grid/item-tile.ts
```diff
@@ -110,47 +115,63 @@ export class ItemTile extends LitElement {
...
   static get styles(): CSSResultGroup {
     return css`
       .container {
         background-color: #ffffff;
         border-radius: var(--collectionTileCornerRadius, 4px);
         box-shadow: 1px 1px 2px 0px;
+        height: 100%;
+      }
+
+      .tile-details {
```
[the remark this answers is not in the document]
dualcnhq: i checked and it still does work. but will update the code still

src/tiles/grid/tile-stats.ts
dualcnhq: i removed `align-items: left` since ideally `text-align: center` should align the numbers stats text

src/tiles/grid/item-tile.ts
```diff
@@ -179,37 +202,36 @@ export class ItemTile extends LitElement {
...
+      span {
         font-size: 1.4rem;
         display: -webkit-box;
       }
 
       h1.truncated {
-        margin-top: 0rem;
-        margin-bottom: 0.5rem;
+        margin: 0px;
         font-size: 1.6rem;
```
nsharma123, line 234: The ticket description says 14px for title and description.

dualcnhq: nice catch, will update it

latonv (approved):

dualcnhq:
thanks for the review @latonv. will merge this now

PR #175, opened 2022-11-28
WEBDEV-5618 Add grid tile on-hover info popups

latonv:
This PR adds small popup panes when hovering over grid tiles, to fulfill the function of the old "collection popups" from legacy search. The popups appear after a 0.3-second delay and show info similar to extended list view (including collections).


src/tiles/hover/hover-pane-controller.ts
```diff
@@ -0,0 +1,463 @@
...
+   * The delay between when the mouse leaves the host element and when the hover
+   * pane should begin fading out (in milliseconds).
+   */
+  private hideDelay: number = 100;
+
+  /**
+   * The delay between when a touch event begins on the host element and when the
+   * hover pane should begin fading in (in milliseconds).
+   */
+  private longPressDelay: number = 600;
```
bfalling, line 101: Nice that you've broken out all these timings into well-documented properties! :)

src/tiles/hover/hover-pane-controller.ts
```diff
@@ -0,0 +1,463 @@
...
+          @touchstart=${this.handleBackdropInteraction}
+          @touchmove=${this.handleBackdropInteraction}
+          @touchend=${this.handleBackdropInteraction}
+          @touchcancel=${this.handleBackdropInteraction}
+          @mouseenter=${(e: MouseEvent) => e.stopPropagation()}
+          @mousemove=${(e: MouseEvent) => e.stopPropagation()}
+          @mouseleave=${(e: MouseEvent) => e.stopPropagation()}
+        ></div>`
+      : nothing;
+  }
```
bfalling, line 196: Nice work on the touchBackdrop, allowing for cancelling the modal by touching _anywhere_ off the pop-up. 👍🏻

src/tiles/hover/hover-pane-controller.ts
```diff
@@ -0,0 +1,463 @@
...
+   * These offsets are only valid if the hover pane is already rendered with its
+   * correct width and height. If the hover pane is not present, the returned offsets
+   * will simply represent the current pointer position.
+   */
+  private get hoverPaneDesiredOffsets(): { top: number; left: number } {
+    // Try to find offsets for the hover pane that:
+    //  (a) cause it to lie entirely within the viewport, and
+    //  (b) to the extent possible, minimize the distance between the
+    //      nearest corner of the hover pane and the mouse position
+    //      (with some additional offsets applied after the fact).
```
bfalling, line 230: This is working so well! :)

src/tiles/tile-dispatcher.ts
```diff
@@ -222,24 +288,45 @@ export class TileDispatcher
...
+        top: 0;
+        left: 0;
+        z-index: 1;
+        background: transparent;
+      }
+
+      tile-hover-pane {
+        position: absolute;
+        top: -2000px;
+        left: -2000px;
```
bfalling, line 327: QUESTION: Is this value intended to take it off screen? If someone has a really big monitor, will these values still work?

bfalling (approved):
Just one question. Otherwise, LGTM!

src/tiles/tile-dispatcher.ts
latonv: That was the intent, yeah -- these positions should generally only appear for a single tick before the popup is positioned correctly, so they're off-screen to avoid a visible flicker. But I bet there's some monitor/situation out there that breaks this assumption. I'll adjust this a bit to future-proof better.

PR #177, opened 2022-12-03
WEBDEV-5654 Ensure uses of infiniteScroller are properly null-guarded

latonv:
The `infiniteScroller` member of collection browser is incorrectly asserted to be non-null. This assertion is invalid in cases where the infinite scroller is not rendered, such as when showing a “0 results” or “no query” placeholder. As a result of the non-null assertion, there are several places in collection-browser.ts that assume `this.infiniteScroller` exists without tsc complaining, producing errors at runtime when that assumption fails (like in the case above).

This PR removes that non-null assertion and ensures uses of `this.infiniteScroller` are optional-chained or null-guarded as appropriate.

### DIFF
--- a/src/collection-browser.ts
+++ b/src/collection-browser.ts
@@ -262,7 +262,7 @@ export class CollectionBrowser
   private dataSource: Record<string, TileModel[]> = {};
 
   @query('infinite-scroller')
-  private infiniteScroller!: InfiniteScroller;
+  private infiniteScroller?: InfiniteScroller;
 
   /**
    * Go to the given page of results
@@ -1142,14 +1142,14 @@ export class CollectionBrowser
     // then scrolls to the cell
     setTimeout(() => {
       this.isScrollingToCell = true;
-      this.infiniteScroller.scrollToCell(cellIndexToScrollTo, true);
+      this.infiniteScroller?.scrollToCell(cellIndexToScrollTo, true);
       // This timeout is to give the scroll animation time to finish
       // then updating the infinite scroller once we're done scrolling
       // There's no scroll animation completion callback so we're
       // giving it 0.5s to finish.
       setTimeout(() => {
         this.isScrollingToCell = false;
-        this.infiniteScroller.reload();
+        this.infiniteScroller?.reload();
       }, 500);
     }, 0);
   }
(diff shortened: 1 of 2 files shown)


src/collection-browser.ts
```diff
@@ -1251,22 +1251,25 @@ export class CollectionBrowser
...
-        this.isScrollingToCell = false;
-        this.infiniteScroller.reload();
-      }, 500);
-    }, 0);
+        this.isScrollingToCell = true;
+        this.infiniteScroller?.scrollToCell(cellIndexToScrollTo, true);
+        // This timeout is to give the scroll animation time to finish
+        // then updating the infinite scroller once we're done scrolling
+        // There's no scroll animation completion callback so we're
+        // giving it 0.5s to finish.
```
iisa, line 1265: q: is animation via css or js? if js, do you think we should add callback on infinite scroller side?

src/collection-browser.ts
```diff
@@ -262,7 +262,7 @@ export class CollectionBrowser
   private dataSource: Record<string, TileModel[]> = {};
 
   @query('infinite-scroller')
-  private infiniteScroller!: InfiniteScroller;
+  private infiniteScroller?: InfiniteScroller;
```
iisa, line 265: > This assertion is invalid in cases where the infinite scroller is not rendered, such as when showing a “0 results” or “no query” placeholder.

interesting... my mental model of collection browser does not make the infinitescroller optional, even without results or without a query started.
Q: what happens if we load empty infinite scroller when there aren't any results?

src/collection-browser.ts
latonv: Hmm, I'm not actually sure -- I just wrapped this existing timeout stuff in a promise for ease of testing. Based on the comment, it sounds like the lack of a callback on infinite scroller may have been intentional, but I'll see if one can be cleanly added.

src/collection-browser.ts
latonv: Good question -- we could probably adjust the templates to keep the infinite scroller rendered to the DOM even when the placeholders replace it visually. Which, I concede, would remove the need for most of these changes 😄. Let me give that a try...

iisa: hi, checking if an empty scroller worked?

latonv: Thanks for the follow-up! I've updated the PR to keep an empty, hidden infinite scroller rendered even when the placeholders are shown. Seems to work fine, and should ensure that the infinite scroller always exists.

iisa (approved):
LGTM TY!

src/collection-browser.ts
```diff
@@ -1363,7 +1375,9 @@ export class CollectionBrowser
     // temporary estimates based on pages rendered so far).
     if (results.length < this.pageSize) {
       this.endOfDataReached = true;
-      this.infiniteScroller.itemCount = this.totalResults;
+      if (this.infiniteScroller) {
```
iisa, line 1378: do we still need conditional if component is there?

PR #178, opened 2022-12-06
New text snippets design 

dualcnhq:


src/tiles/text-snippet-block.ts
```diff
@@ -90,32 +82,36 @@ export class TextSnippetBlock extends LitElement {
   static get styles(): CSSResultGroup {
     return css`
       .container {
+        display: flex;
+        flex-direction: row;
+        flex-wrap: wrap;
+        width: 100%;
+        border-left: 5px solid #194880;
```
latonv, line 89: Can we add a 3px border-radius too? (I think this was accidentally omitted from the spec document, but it's present in the images).

latonv (approved):
LGTM, with two minor suggestions.

src/tiles/text-snippet-block.ts
dualcnhq: that's weird because in current offshoot prod, adding it back or removing it doesn't do anything 🤔

dualcnhq: ohhhh... i notice it now, it's in the border-left corners

dualcnhq: updated now

PR #180, opened 2022-12-08
WEBDEV-5281 Fix year histogram interactions & consolidate requests

latonv:
The date picker's handle interactions have unusual offsets from the pointer position on the beta search page. These interactions have been fixed in the version of the new version of the date picker component that this PR upgrades to.

Also, since the PPS now provides correct year histogram data alongside other aggregations regardless of the state of other year filters, a separate request for it is no longer necessary. Accordingly, this PR removes that separate request and several now-obsolete collection browser methods that supported it.


iisa (approved):
nice sweep

PR #185, opened 2022-12-14
WEBDEV-5680 Ensure title/creator letter filters clear when mobile sort field changes

latonv:
Resolves a bug in mobile view, where selecting a different sort option would still leave any existing title/creator letter filter active.

### DIFF
--- a/test/sort-filter-bar/sort-filter-bar.test.ts
+++ b/test/sort-filter-bar/sort-filter-bar.test.ts
@@ -229,3 +229,92 @@ describe('Sort/filter bar letter behavior', () => {
     expect(el.selectedCreatorFilter).to.equal('C');
   });
 });
+
+describe('Sort/filter bar mobile view', () => {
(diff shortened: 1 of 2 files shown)


test/sort-filter-bar/sort-filter-bar.test.ts
```diff
@@ -229,3 +229,92 @@ describe('Sort/filter bar letter behavior', () => {
...
+describe('Sort/filter bar mobile view', () => {
+  let origWindowSize: { width: number; height: number };
+  before(() => {
+    origWindowSize = { width: window.innerWidth, height: window.innerHeight };
+    window.resizeTo(500, 600);
+  });
+
+  after(() => {
+    window.resizeTo(origWindowSize.width, origWindowSize.height);
+  });
```
iisa, line 242: nice use of before/after

iisa (approved):
nice tests :nice:  :lgtm:

PR #186, opened 2022-12-16
WEBDEV-5690 Adjust grid tile layout

latonv:
Removes some of the whitespace between lines on grid tiles, and ensures the stats row at the bottom is aligned consistently.

### DIFF
--- a/src/tiles/grid/styles/tile-grid-shared-styles.ts
+++ b/src/tiles/grid/styles/tile-grid-shared-styles.ts
@@ -21,7 +21,6 @@ export const baseTileStyles = css`
 
   image-block {
     display: block;
-    margin-bottom: 5px;
     position: relative;
     text-align: center;
   }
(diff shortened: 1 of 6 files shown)


src/tiles/grid/tile-stats.ts
```diff
@@ -43,7 +43,7 @@ export class TileStats extends LitElement {
             <p class="sr-only">Mediatype:</p>
             <mediatype-icon .mediatype=${this.mediatype}></mediatype-icon>
           </li>
-          <li class="col" title="${uploadsOrViewsTitle}">
+          <li class="col views" title="${uploadsOrViewsTitle}">
```
dualcnhq, line 46: views and favorites class is missing here, only reviews is declared in the CSS styles.
can we come up with a more generic class name for this instead?

src/tiles/grid/styles/tile-grid-shared-styles.ts
```diff
@@ -35,14 +34,14 @@ export const baseTileStyles = css`
   }
 
   .item-info {
+    display: flex;
+    flex-direction: column;
```
dualcnhq, line 38: :nice:

src/tiles/grid/tile-stats.ts
latonv: We can certainly remove the unused "views" and "favorites" classes (unless we need to make different adjustments to those icons too).

For the reviews icon though, I am _specifically_ targeting that one for a tailored adjustment that likely doesn't generalize elsewhere...

src/tiles/grid/styles/tile-grid-shared-styles.ts
latonv: Flexbox is great 😄

PR #187, opened 2022-12-17
WEBDEV-5695 Correct compact list view headers & date column defaults

latonv:
In the compact list view:
- Adds a "Type" column header to the column with the mediatype icons (since it looks strange without any header)
- Defaults the date column to show date published, not date archived (unless there is a date sort active, which takes precedence).

### DIFF
--- a/src/tiles/list/tile-list-compact.ts
+++ b/src/tiles/list/tile-list-compact.ts
@@ -66,6 +66,12 @@ export class TileListCompact extends LitElement {
    * @see src/models.ts
    */
   private get date(): Date | undefined {
+    // Note on 'publicdate' vs. 'date':
+    // The search engine metadata uses 'publicdate' as the key for the date the item
+    // was created on archive.org, which in the UI is referred to as "Date Archived".
+    // In contrast, the search engine metadata uses 'date' to refer to the actual
+    // publication date of the underlying media ("Date Published" in the UI).
+    // Refer to the full metadata schema for more info.
     switch (this.sortParam?.field) {
       case 'publicdate':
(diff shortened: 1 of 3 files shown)


src/tiles/list/tile-list-compact.ts
```diff
@@ -67,14 +67,14 @@ export class TileListCompact extends LitElement {
...
-        return this.model?.datePublished;
+      case 'publicdate':
+        return this.model?.dateArchived;
       case 'reviewdate':
         return this.model?.dateReviewed;
       case 'addeddate':
         return this.model?.dateAdded;
       default:
-        return this.model?.dateArchived; // publicdate
+        return this.model?.datePublished; // 'date'
```
dualcnhq, line 77: didn't quite understand the comment here, would like to suggest to remove the comment instead

dualcnhq (approved):
LGTM, just 1 thing am concerned about but not a blocker

src/tiles/list/tile-list-compact.ts
```diff
@@ -66,15 +66,21 @@ export class TileListCompact extends LitElement {
    * @see src/models.ts
    */
   private get date(): Date | undefined {
+    // Note on 'publicdate' vs. 'date':
+    // The search engine metadata uses 'publicdate' as the key for the date the item
+    // was created on archive.org, which in the UI is referred to as "Date Archived".
+    // In contrast, the search engine metadata uses 'date' to refer to the actual
+    // publication date of the underlying media ("Date Published" in the UI).
+    // Refer to the full metadata schema for more info.
```
dualcnhq, line 74: 👍

PR #189, opened 2022-12-22
WEBDEV-5686 Ensure grid tiles are not overly wide, fit 2 cols in mobile view

latonv:
Currently, at mobile device portrait view widths, often there is only one column of grid tiles shown, and the tiles appear overly wide.

This PR reduces the minimum widths of tiles at different breakpoints to ensure that they can continue to fit 2 columns even down to very narrow device widths. These smaller min widths also allow the tile columns to shrink further before a column is removed, which has the side effect of reducing the maximum tile width possible.

### DIFF
--- a/src/collection-browser.ts
+++ b/src/collection-browser.ts
@@ -1700,11 +1700,20 @@ export class CollectionBrowser
     infinite-scroller.grid {
       --infiniteScrollerCellMinWidth: var(
         --collectionBrowserCellMinWidth,
-        18rem
+        17rem
       );
       --infiniteScrollerCellMaxWidth: var(--collectionBrowserCellMaxWidth, 1fr);
     }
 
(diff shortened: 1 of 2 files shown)


src/collection-browser.ts
```diff
@@ -1700,11 +1700,30 @@ export class CollectionBrowser
...
       --infiniteScrollerCellMinWidth: var(
         --collectionBrowserCellMinWidth,
-        18rem
+        17rem
       );
       --infiniteScrollerCellMaxWidth: var(--collectionBrowserCellMaxWidth, 1fr);
     }
 
+    /* Allow tiles to shrink a bit further at smaller viewport widths */
+    @media screen and (max-width: 880px) {
```
iisa, line 1709: mixing a couple of my favorite css tricks: media queries with global vars! :nice:

iisa (approved):
🚢

PR #190, opened 2022-12-22
WEBDEV-5693 Adjust extended list view thumbnail styles for better visibility/alignment

latonv:
The container box around extended list view thumbnails blends in very closely with the background on archive.org, making it look like the aspect-fit thumbnails are misaligned when they are wider than they are tall. This PR adjusts the styles for the container box to make the box more visibly aligned with the title row.

Also gives the demo app the same color background as archive.org, to serve as a better point of comparison.

### DIFF
--- a/src/tiles/image-block.ts
+++ b/src/tiles/image-block.ts
@@ -106,6 +106,12 @@ export class ImageBlock extends LitElement {
       }
 
       /** tile-list view */
+      .list {
+        border-radius: 0;
+        background-color: var(--imageBlockListBackgroundColor, #ebebee);
+        box-shadow: 1px 1px 2px rgb(0, 0, 0, 0.2);
+      }
+
       .list.desktop {
         width: 100px;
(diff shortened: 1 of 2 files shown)


src/tiles/image-block.ts
```diff
@@ -106,6 +106,12 @@ export class ImageBlock extends LitElement {
       }
 
       /** tile-list view */
+      .list {
+        border-radius: 0;
+        background-color: var(--imageBlockListBackgroundColor, #ebebee);
```
iisa, line 111: `#ebebee` is new to me! works for me if @pezvi approved 👍

iisa (approved):

src/tiles/image-block.ts
latonv: Oh yeah, I meant to check if there's a better color already in use that would fit here. I'll talk to @pezvi & adjust accordingly. Thanks! 🐝

PR #191, opened 2022-12-22
WEBDEV-5671 Remove facet sort toggles from sidebar, and convert to links in More dialog

latonv:
The little sorting toggle buttons in the facet sidebar are confusing and unclear. There is already a means of sorting the facets within the More dialog, so the minor shortcut these buttons provide is not particularly helpful.

This PR removes those toggle buttons from the sidebar, and adjusts the sorting affordance in the More dialog to use a textual label describing the sort rather than the previous button's unclear visuals.

### DIFF
--- a/src/collection-facets/more-facets-content.ts
+++ b/src/collection-facets/more-facets-content.ts
@@ -397,26 +397,21 @@ export class MoreFacetsContent extends LitElement {
   }
 
   private get getModalHeaderTemplate(): TemplateResult {
-    const title =
-      this.sortedBy === 'alpha' ? 'Sort by count' : 'Sort by alphabetically';
-
-    const image =
-      this.sortedBy === 'alpha'
-        ? 'https://archive.org/images/filter-alpha.png'
-        : 'https://archive.org/images/filter-count.png';
+    const title = this.sortedBy === 'alpha' ? 'Sort by count' : 'Sort by value';
 
     return html`<span class="sr-only">More facets for:</span>
       <span class="title">
         ${this.facetGroupTitle}
-        <input
-          class="sorting-icon"
-          type="image"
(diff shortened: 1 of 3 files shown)


src/collection-facets/more-facets-content.ts
```diff
@@ -397,26 +397,21 @@ export class MoreFacetsContent extends LitElement {
...
-        <input
-          class="sorting-icon"
-          type="image"
-          @click=${() => this.sortFacetAggregation()}
-          src="${image}"
-          title=${title}
-          alt="sort facets"
-        />
-      </span> `;
+        <a
```
iisa, line 405: q: can we use button since it isn't really linking out to anywhere?

src/collection-facets/more-facets-content.ts
```diff
@@ -481,6 +476,16 @@ export class MoreFacetsContent extends LitElement {
         padding: 0 10px;
         font-weight: bold;
       }
+      .sort-link {
+        margin-left: 0.7rem;
+        color: var(--ia-theme-link-color, #4b64ff);
```
iisa, line 481: love a good default 👍

iisa (changes requested):
just a question about using button vs anchor.

I really like twill's write up about the distinction between the two: https://paste.twilio.design/patterns/button-vs-anchor#general

```
From a high level, buttons are to be used for performing an in-page action and anchors/links are for navigating to a new page.
```

iisa (approved):
once element update is in, lgtm 👍 thanks!!

src/collection-facets/more-facets-content.ts
latonv: Thanks for the suggestion! Agreed a button is more semantically accurate here -- updated.

PR #188, opened 2022-12-20
WEBDEV-5694 Make mediatype icons in list view link to their collection

latonv:
Converts extended list view mediatype icons into links targeting their mediatype's collection details page.

PR #196, opened 2022-12-23
WEBDEV-5717 Allow mediatype links with empty baseNavigationUrl

latonv:
Patches an issue with #188 where the links were not generating correctly in the production environment where `baseNavigationUrl` is an empty string.


bfalling (approved):
LGTM

PR #200, opened 2022-12-29
WEBDEV-5707 Allow info popups to be toggled by tapping info icon

latonv:
The info popups that appear when hovering over grid tiles are currently not available to mobile/touch device users that don't have hover capability. To present an alternative for these users, this PR adds a small info icon in the corner of grid tiles when hover interactions are not available, which the user can tap to show/hide the info popup for that tile.

### DIFF
--- a/src/tiles/grid/account-tile.ts
+++ b/src/tiles/grid/account-tile.ts
@@ -1,4 +1,4 @@
-import { css, html, LitElement, TemplateResult } from 'lit';
+import { css, html, LitElement, nothing, TemplateResult } from 'lit';
 import { customElement, property } from 'lit/decorators.js';
 import type { TileModel } from '../../models';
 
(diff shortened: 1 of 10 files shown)


src/tiles/grid/account-tile.ts
```diff
@@ -59,6 +62,25 @@ export class AccountTile extends LitElement {
     </tile-stats>`;
   }
 
+  private get infoButtonTemplate(): TemplateResult | typeof nothing {
+    // &#9432; is an information icon
```
dualcnhq, line 66: 👍

src/tiles/grid/collection-tile.ts
```diff
@@ -56,7 +66,7 @@ export class CollectionTile extends LitElement {
   }
 
   private get getItemsTemplate() {
-    const collectionItems = this.model?.itemCount.toLocaleString();
+    const collectionItems = this.model?.itemCount?.toLocaleString();
 
     return html`<span id="item-count"
```
dualcnhq, line 71: some weird code formatting here 🤔

dualcnhq (approved):

PR #205, opened 2023-01-04
WEBDEV-5737 Use PPS-provided href for tile links if available

latonv:
On legacy search, FTS tiles include the current search query as a parameter when linking to items' details pages. This allows the theatre on the details page to perform its search-inside function when loaded. However, currently on the beta search page, FTS results don’t include this parameter.

The PPS provides an href field on FTS hits that includes the full link target, including the query string. Moreover, this field is intended to hold other contextual information (like start/end times) for TV/Radio hits when those are eventually enabled. So where possible, we should be making use of this field instead of constructing details page links manually.

This PR ensures that for FTS results (and other future result types), the tile links use the PPS href field that includes additional context params.

### DIFF
--- a/src/tiles/tile-dispatcher.ts
+++ b/src/tiles/tile-dispatcher.ts
@@ -122,7 +122,7 @@ export class TileDispatcher
   private get linkTileTemplate() {
     return html`
       <a
-        href="${this.baseNavigationUrl}/details/${this.model?.identifier}"
+        href="${this.linkTileHref}"
         title=${this.shouldPrepareHoverPane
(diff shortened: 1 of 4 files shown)


src/tiles/tile-dispatcher.ts
```diff
@@ -136,6 +136,15 @@ export class TileDispatcher
     `;
   }
 
+  private get linkTileHref() {
```
dualcnhq, line 139: 👍

dualcnhq (approved):

PR #207, opened 2023-01-06
WEBDEV-5287 Implement revised sort bar design

latonv:
This PR implements the new sort bar design that, most importantly:
- uses custom dropdown menus instead of pill buttons for selecting view/date sorting options
- has a clearer/easier-to-use sort direction toggle
- applies more sensible default sort directions for each sorting option
- removes the large padding and box-shadow from the sort bar for a more streamlined look

Included are a few smaller design tweaks:
- fixing the colors of the display mode options in their selected/unselected states
- normalizing the order in which sorting options are displayed across desktop/mobile
- sentence-casing the sorting option labels
- ensuring the correct view type (weekly/all-time) is shown on grid tiles depending on the sort


src/models.ts
```diff
@@ -79,16 +80,30 @@ export const SortFieldDisplayName: {
...
   title: 'Title',
-  datearchived: 'Date Archived',
-  date: 'Date Published',
-  datereviewed: 'Date Reviewed',
-  dateadded: 'Date Added',
+  date: 'Date published',
+  datearchived: 'Date archived',
+  datereviewed: 'Date reviewed',
+  dateadded: 'Date added',
   creator: 'Creator',
```
latonv, line 90: (Switching from Title Case to Sentence case)

src/models.ts
```diff
@@ -99,8 +114,8 @@ export const SortFieldToMetadataField: {
   alltimeview: 'downloads',
   weeklyview: 'week',
   title: 'titleSorter',
-  datearchived: 'publicdate',
   date: 'date',
+  datearchived: 'publicdate',
```
latonv, line 118: (Just making the order of these consistent)

src/sort-filter-bar/img/list.ts
```diff
@@ -1,5 +1,5 @@
...
...83915 1.5611039-.8706412h65.7343743c.5600372 0 1.0508477.2902137 1.4724313.8706412.4215836.5804274.6315976 1.2755905.6300505 2.0854893 0 .8683914-.2100226 1.5928009-.6300505 2.1732283-.420028.5804275-.9108384.8706412-1.4724313.8706412z"/><path d="m97.8975061 45h-65.7343743c-.6409315 0-1.1612995-.2902137-1.5611039-.8706412-.3998043-.5804274-.6004844-1.3048369-.60204-2.1732283 0-.8121485.20068-1.5073116.60204-2.0854893.4013601-.5781778.921728-.8683915 1.5611039-.8706412h65.7343743c.5600372 0 1.0508477.2902137 1.4724313.8706412.4215836.5804274.6315976 1.2755905.6300505 2.0854893 0 .8683914-.2100226 1.5928009-.6300505 2.1732283-.420028.5804275-.9108384.8706412-1.4724313.8706412z"/><path d="m97.8975061 100h-65.7343743c-.6409315 0-1.1612995-.2902137-1.5611039-.8706412-.3998043-.5804274-.6004844-1.3048369-.60204-2.1732283 0-.8121485.20068-1.5073116.60204-2.0854893.4013601-.5781778.921728-.8683915 1.5611039-.8706412h65.7343743c.5600372 0 1.0508477.2902137 1.4724313.8706412.4215836.5804274.6315976 1.2755905.6300505 2.0854893 0 .8683914-.2100226 1.5928009-.6300505 2.1732283-.420028.5804275-.9108384.8706412-1.4724313.8706412z"/><path d="m0 0h25v25h-25z"/><path d="m0 55h25v25h-25z"/></g></svg>
```
latonv, line 4: (This svg erroneously had a fill set on it -- removing so we can control the fill from CSS)

src/sort-filter-bar/sort-filter-bar.ts
```diff
@@ -205,110 +209,147 @@ export class SortFilterBar
...
-        >
-          ${sortIcon}
-        </button>
+      <button
+        class="sort-direction-selector"
+        ?disabled=${this.selectedSort === 'relevance'}
+        @click=${this.toggleSortDirection}
+      >
+        ${this.sortDirectionIcon}
+      </button>
```
latonv, line 227: The sort direction selector is now a single button that displays the whole up/down arrow icon, rather than two separate triangle buttons.

src/sort-filter-bar/sort-filter-bar.ts
```diff
@@ -180,15 +221,31 @@ export class SortFilterBar
     });
 
     this.resizeObserver.addObserver({
-      target: this.desktopSortSelector,
+      target: this.desktopSortContainer,
       handler: this,
     });
   }
 
+  handleResize(entry: ResizeObserverEntry): void {
```
latonv, line 229: (This is just moved up from below)

src/sort-filter-bar/sort-filter-bar.ts
```diff
@@ -204,122 +261,112 @@ export class SortFilterBar
...
-          })}
-        </li>
-      </ul>
+        <ul id="desktop-sort-selector">
+          <li>
+            ${this.showRelevance
+              ? this.getSortDisplayOption(SortField.relevance)
+              : nothing}
+          </li>
+          <li>${this.viewsDropdownTemplate}</li>
```
latonv, line 305: Changed views/date to use dropdown components instead of the old "pill" selector popups. I've tried to keep things readable by splitting out the dropdown templates below.

src/tiles/grid/item-tile.ts
```diff
@@ -35,6 +35,10 @@ export class ItemTile extends LitElement {
 
   render() {
     const itemTitle = this.model?.title;
+    const [viewCount, viewLabel] =
```
latonv, line 38: This isn't _strictly_ part of the sort bar, but it fixes the grid tiles so they show the correct type of views depending on the selected sort option.

src/models.ts
```diff
@@ -55,8 +56,8 @@ export enum SortField {
   'alltimeview' = 'alltimeview',
   'weeklyview' = 'weeklyview',
   'title' = 'title',
-  'datearchived' = 'datearchived',
```
latonv, line 58: The order of these should be consistent, as it is how they appear in the mobile-view dropdown.

src/models.ts
```diff
@@ -55,8 +56,8 @@ export enum SortField {
...
   'date' = 'date',
+  'datearchived' = 'datearchived',
   'datereviewed' = 'datereviewed',
   'dateadded' = 'dateadded',
   'creator' = 'creator',
```
latonv, line 63: The order of these should be consistent, as it is how they appear in the mobile-view dropdown.

src/models.ts
bfalling: Funny. Never heard it called "Sentence Case" (but makes total sense!).

src/tiles/grid/item-tile.ts
```diff
@@ -35,6 +35,10 @@ export class ItemTile extends LitElement {
...
+      this.sortParam?.field === 'week'
+        ? [this.model?.weeklyViewCount, 'weekly views']
+        : [this.model?.viewCount, 'all-time-views'];
```
bfalling, line 41: QUESTION: Inconsistent value formatting? Should there be a space before the word "views" in both cases?

src/sort-filter-bar/sort-filter-bar.ts
```diff
@@ -204,122 +255,120 @@ export class SortFilterBar
...
+          dropdownOptions: Object.keys(SortField).map(field =>
+            this.getDropdownOption(field as SortField)
+          ),
+          selectedOption: this.selectedSort ?? SortField.relevance,
+          onOptionSelected: this.mobileSortChanged,
+          onDropdownClick: () => {
+            this.dropdownBackdropVisible = this.mobileDropdown.open;
+            this.mobileDropdown.classList.toggle(
+              'open',
+              this.mobileDropdown.open
```
bfalling, line 356: SUGGESTION: Add a comma at the end of this line...?

src/sort-filter-bar/sort-filter-bar.ts
```diff
@@ -341,46 +390,148 @@ export class SortFilterBar
...
-        class=${this.mobileSelectorVisible ? 'visible' : 'hidden'}
+      <ia-dropdown
+        id=${options.id ?? nothing}
+        class=${options.isSelected?.() ? 'selected' : nothing}
+        displayCaret
+        closeOnSelect
+        includeSelectedOption
+        .openViaButton=${false}
+        .options=${options.dropdownOptions}
+        .selectedOption=${options.selectedOption}
```
bfalling, line 435: SUGGESTION: `selectedOption` could be null, right? So provide default with `??`?

src/sort-filter-bar/sort-filter-bar.ts
```diff
@@ -341,46 +390,148 @@ export class SortFilterBar
...
+      selectedOption: this.viewOptionSelected ? this.selectedSort : '',
+      onOptionSelected: this.dropdownOptionSelected,
+      onDropdownClick: () => {
+        this.dateDropdown.open = false;
+        this.dropdownBackdropVisible = this.viewsDropdown.open;
+        this.viewsDropdown.classList.toggle('open', this.viewsDropdown.open);
+      },
+      onLabelInteraction: () => {
+        if (!this.viewsDropdown.open && !this.viewOptionSelected)
+          this.setSelectedSort(SortField.weeklyview);
```
bfalling, line 499: QUESTION: Do we generally allow one-line `if` bodies with no braces? On two separate lines? (Same question applies below.)

bfalling (commented):
One possible bug, but otherwise, no blockers. A bunch of suggestions and questions.

src/tiles/grid/item-tile.ts
latonv: Ah, good catch! Typo should be fixed in the latest commit.

src/sort-filter-bar/sort-filter-bar.ts
latonv: Our Prettier ruleset doesn't allow trailing commas on argument lists (only on multiline arrays/objects). This is probably fine since it would be an error to add additional arguments here.

src/sort-filter-bar/sort-filter-bar.ts
latonv: Thanks, should be fixed now!

src/sort-filter-bar/sort-filter-bar.ts
latonv: Thanks, I've added braces around this and the similar occurrences below.

PR #213, opened 2023-01-12
WEBDEV-5763 Allow links on extended & compact list view to use visited styling

latonv:
To support users who rely on knowing which links they have visited from a search page, we should allow links to present their “visited” styling rather than forcing the blue color in all cases.

### DIFF
--- a/src/tiles/list/tile-list-compact.ts
+++ b/src/tiles/list/tile-list-compact.ts
@@ -66,7 +66,7 @@ export class TileListCompact extends LitElement {
     // Use the server-specified href if available.
     // Otherwise, construct a details page URL from the item identifier.
     return this.model?.href
-      ? `${this.baseNavigationUrl}${this.model?.href}`
+      ? `${this.baseNavigationUrl}${this.model.href}`
       : `${this.baseNavigationUrl}/details/${this.model?.identifier}`;
   }
 
--- a/src/tiles/list/tile-list-compact.ts
+++ b/src/tiles/list/tile-list-compact.ts
@@ -41,7 +41,9 @@ export class TileListCompact extends LitElement {
           .loggedIn=${this.loggedIn}
         >
         </image-block>
-        <div id="title">${DOMPurify.sanitize(this.model?.title ?? '')}</div>
+        <a href=${this.href} id="title"
+          >${DOMPurify.sanitize(this.model?.title ?? '')}</a
+        >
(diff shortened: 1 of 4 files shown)


iisa (approved):

src/tiles/list/tile-list-compact.ts
```diff
@@ -41,7 +41,9 @@ export class TileListCompact extends LitElement {
           .loggedIn=${this.loggedIn}
         >
         </image-block>
-        <div id="title">${DOMPurify.sanitize(this.model?.title ?? '')}</div>
+        <a href=${this.href} id="title"
```
bfalling, line 44: QUESTION: This "id" isn't used anywhere else, is it?

src/tiles/list/tile-list-compact.ts
```diff
@@ -60,6 +62,14 @@ export class TileListCompact extends LitElement {
     `;
   }
 
+  private get href(): string {
+    // Use the server-specified href if available.
+    // Otherwise, construct a details page URL from the item identifier.
+    return this.model?.href
+      ? `${this.baseNavigationUrl}${this.model?.href}`
+      : `${this.baseNavigationUrl}/details/${this.model?.identifier}`;
```
bfalling, line 70: QUESTION: Is there a way to test both of these code pathways?

bfalling (approved):
Two questions, non-blockers probably. Otherwise, LGTM!

src/tiles/list/tile-list-compact.ts
latonv: Currently, full-text searches receive an `href` from the PPS, while metadata searches do not. So full-text should follow the top branch, while metadata should follow the bottom branch.

In practice, the only difference between the two is that FTS links append the current search query so that the search inside pane in BookReader opens automatically when you visit the item. However, in the future, TV and Radio searches will also provide custom `href` fields with things like start/end times set.

src/tiles/list/tile-list-compact.ts
latonv: It's used in the CSS below to style this element, but I think that's it.

latonv: Oh, on a re-read, I think you might have been asking if it could collide with another `id` somewhere. Short answer is it's scoped to this component's shadow DOM, so shouldn't be an issue.

PR #218, opened 2023-01-20
WEBDEV-5785 Prevent query mismatch via uid param (and trim queries)

latonv:
Search queries with leading or trailing spaces are currently being sent to the backend as-is. Although the backend trims the queries anyway, the response we get back consequently has a different query string than the one we sent. Currently, the tile renderer treats this query mismatch by ignoring what it assumes is an obsolete response to an old query, rendering nothing.

While the simple fix is just to trim the query, we can go one step further and mitigate future “query mismatch” issues of this nature by sending a `uid` parameter with each search request to the PPS. This parameter is not processed and should always be returned as-is on the response. If we maintain a consistent `uid` for queries that should be considered the same, then we can use this as a much better signal of response-freshness than matching query strings (which the server may alter, as in this case).

This PR updates the page & facet fetch logic to use the `uid` parameter as described above, and ensures that queries are trimmed before sending to the PPS.

### DIFF
--- a/src/collection-browser.ts
+++ b/src/collection-browser.ts
@@ -1252,8 +1252,8 @@ export class CollectionBrowser
    * no longer relevant.
    */
   private get pageFetchQueryKey(): string {
-    const sortField = this.sortParam?.field ?? '';
-    const sortDirection = this.sortParam?.direction ?? '';
+    const sortField = this.sortParam?.field ?? 'none';
+    const sortDirection = this.sortParam?.direction ?? 'none';
     return `${this.fullQuery}-${this.searchType}-${sortField}-${sortDirection}`;
   }
 
--- a/src/collection-browser.ts
+++ b/src/collection-browser.ts
@@ -345,7 +345,7 @@ export class CollectionBrowser
 
   private setPlaceholderType() {
     this.placeholderType = null;
-    if (!this.baseQuery) {
+    if (!this.baseQuery?.trim()) {
       this.placeholderType = 'empty-query';
     }
 
@@ -1008,7 +1008,7 @@ export class CollectionBrowser
   /** The base query joined with any title/creator letter filters */
   private get filteredQuery(): string | undefined {
     if (!this.baseQuery) return undefined;
-    let filteredQuery = this.baseQuery;
+    let filteredQuery = this.baseQuery.trim();
 
     const { sortFilterQueries } = this;
     if (sortFilterQueries) {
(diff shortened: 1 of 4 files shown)


src/collection-browser.ts
```diff
@@ -1117,18 +1117,38 @@ export class CollectionBrowser
...
+    const returnedUid = (success.request.clientParameters as any).uid;
+    const queryChangedSinceFetch = returnedUid !== this.facetFetchQueryKey;
+    if (queryChangedSinceFetch) return;
+
+    this.aggregations = success?.response.aggregations;
 
     this.fullYearsHistogramAggregation =
-      results?.success?.response?.aggregations?.year_histogram ??
-      results?.success?.response?.aggregations?.['year-histogram']; // Temp fix until PPS FTS key is fixed to use underscore
+      success?.response?.aggregations?.year_histogram;
```
latonv, line 1151: The PPS bug that necessitated this workaround is no longer present. 🎉

src/collection-browser.ts
```diff
@@ -1165,8 +1185,18 @@ export class CollectionBrowser
    * This lets us keep track of queries so we don't persist data that's
    * no longer relevant.
    */
-  private get pageFetchQueryKey() {
-    return `${this.fullQuery}-${this.searchType}-${this.sortParam?.field}-${this.sortParam?.direction}`;
+  private get pageFetchQueryKey(): string {
+    const sortField = this.sortParam?.field ?? '';
+    const sortDirection = this.sortParam?.direction ?? '';
+    return `${this.fullQuery}-${this.searchType}-${sortField}-${sortDirection}`;
```
iisa, line 1191: Q: ok if returned string is `foo-bar--`?

src/collection-browser.ts
```diff
@@ -954,7 +954,7 @@ export class CollectionBrowser
       filteredQuery += ` AND ${sortFilterQueries}`;
     }
 
-    return filteredQuery;
+    return filteredQuery.trim();
```
iisa, line 957: q: can/should we trim filteredQuery at declaration?

iisa (commented):
looks good.
chunks changing, what are good ways to test the trim + other updates?

src/collection-browser.ts
latonv: Ok and expected. Though I suppose an alternative would be to put a special `default` value in for the sort field/direction when they aren't defined.

src/collection-browser.ts
latonv: Yes, in fact defensively trimming it in both places may be sensible (trim base query at declaration, and trim again after adding filters). Updated + tests added!

src/collection-browser.ts
latonv: On second thought, it's better if these more explicitly convey something. So the no-sort case should be `foo-bar-none-none` now.

PR #219, opened 2023-01-24
WEBDEV-5796 Ensure sort direction icon svg is properly sized on all supported browsers

latonv:
For some browsers (esp. on Mac/iOS), the sort direction icons are rendering at their intrinsic size rather than fitting within the available space, causing them to not be visible. This PR restricts the height of those icon svgs to that of the sort bar, preventing the issue.

### DIFF
--- a/src/sort-filter-bar/sort-filter-bar.ts
+++ b/src/sort-filter-bar/sort-filter-bar.ts
@@ -851,6 +851,11 @@ export class SortFilterBar
       height: 14px;
     }
 
+    .sort-direction-icon > svg {
+      height: -webkit-fit-content;
+      height: fit-content;
+    }
+
     #date-sort-selector,
     #view-sort-selector {
       position: absolute;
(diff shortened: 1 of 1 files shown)


src/sort-filter-bar/sort-filter-bar.ts
```diff
@@ -851,6 +851,11 @@ export class SortFilterBar
       height: 14px;
     }
 
+    .sort-direction-icon > svg {
+      height: -webkit-fit-content;
+      height: fit-content;
```
cdrini, line 856: I found `flex: 1` seemed to do the trick as well on iPad + safari ; that might be a cleaner approach? But the provided does work, so whatever works!

cdrini (approved):
Lgtm! 1 possible suggestion.

src/sort-filter-bar/sort-filter-bar.ts
latonv: Oh, you're right! That does seem more straightforward.

PR #220, opened 2023-01-24
WEBDEV-5800 Fix 3-line truncation of list tiles in Safari

latonv:
Several blocks of metadata on list view tiles (esp. the description and full text snippets) are meant to be truncated after the first 3 lines. While this appears to be working across other browsers, Safari is not truncating them at all, which leads to excessively long walls of text in hover panes.

This PR introduces a wrapper element around these blocks of text that makes Safari truncate them properly.

### DIFF
--- a/src/tiles/list/tile-list.ts
+++ b/src/tiles/list/tile-list.ts
@@ -305,7 +305,7 @@ export class TileList extends LitElement {
     if (!text) return nothing;
     return html`
       <div id=${ifDefined(id)} class="metadata">
-        ${this.labelTemplate(label)} ${text}
+        <p class="inline-wrap">${this.labelTemplate(label)} ${text}</p>
       </div>
     `;
   }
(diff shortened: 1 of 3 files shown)


src/tiles/list/tile-list.ts
```diff
@@ -305,7 +305,7 @@ export class TileList extends LitElement {
     if (!text) return nothing;
     return html`
       <div id=${ifDefined(id)} class="metadata">
-        ${this.labelTemplate(label)} ${text}
+        <p class="inline-wrap">${this.labelTemplate(label)} ${text}</p>
```
cdrini, line 308: Do we need the extra element? I found adding `max-height` to the `<div>` did the trick as well. A little less nesting :)

cdrini (approved):
Lgtm! 1 possible suggestion to reduce some of the HTML depth

src/tiles/list/tile-list.ts
latonv: These elements were originally added because Safari was entirely ignoring the line-clamp rule when the child wasn't a single inline element. But yeah, not necessary anymore with the `max-height` clamping. 🙂

PR #222, opened 2023-01-25
WEBDEV-5804 Correct handling of `sin` param

latonv:
This PR simplifies how the search types are provided to collection browser from its parent component, so that requesting an additional search is no longer required. Search type changes due to history stack pops now correctly propagate upwards (e.g., to set the correct radio button option on the search page).

### DIFF
--- a/src/app-root.ts
+++ b/src/app-root.ts
@@ -90,10 +90,6 @@ export class AppRoot extends LitElement {
     this.searchQuery = this.baseQueryField.value;
     this.collectionBrowser.searchType = this.searchType;
 
-    // Need to request a search in cases where the search type was previously changed but not the query.
-    // If the query was changed above, this should be a no-op.
-    this.collectionBrowser.requestSearch();
-
     if ((this.currentPage ?? 1) > 1) {
       this.collectionBrowser.goToPage(this.currentPage ?? 1);
     }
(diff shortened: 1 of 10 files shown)


src/app-root.ts
```diff
@@ -90,10 +90,6 @@ export class AppRoot extends LitElement {
     this.searchQuery = this.baseQueryField.value;
     this.collectionBrowser.searchType = this.searchType;
 
-    // Need to request a search in cases where the search type was previously changed but not the query.
-    // If the query was changed above, this should be a no-op.
-    this.collectionBrowser.requestSearch();
-
```
latonv, line 96: Removed Offshoot's reliance on this, so bringing the demo app in line with the same flow.

src/collection-browser.ts
```diff
@@ -321,14 +321,6 @@ export class CollectionBrowser
...
   }
 
-  /**
-   * Manually requests to perform a search, which will only be executed if one of
-   * the query, the search type, or the sort has changed.
-   */
-  requestSearch() {
-    this.handleQueryChange();
-  }
-
```
latonv, line 331: No longer used/needed.

src/restoration-state-handler.ts
```diff
@@ -233,9 +233,14 @@ export class RestorationStateHandler
...
-      restorationState.searchType =
-        searchInside === 'TXT' ? SearchType.FULLTEXT : SearchType.METADATA;
+    switch (searchInside) {
+      // Eventually there will be TV/Radio search types here too.
+      case 'TXT':
+        restorationState.searchType = SearchType.FULLTEXT;
+        break;
+      default:
+        restorationState.searchType = SearchType.METADATA;
+        break;
```
latonv, line 243: Empty or invalid `sin` params should just default to metadata search type.

iisa (approved):

PR #228, opened 2023-01-31
WEBDEV-5637 Ensure alphabet filter counts load when present on initial page load

latonv:
This PR fixes a bug that prevented the alphabet bar's bucket counts from loading correctly when they were part of the initial query on page load (e.g., via a `firstTitle` or `firstCreator` filter in the URL). It also updates the requests for these counts to properly use the search service's filter map instead of joining `AND` clauses to the base query.

### DIFF
--- a/src/collection-browser.ts
+++ b/src/collection-browser.ts
@@ -1162,10 +1162,12 @@ export class CollectionBrowser
   }
 
   private async fetchFacets() {
-    if (!this.filteredQuery) return;
+    const trimmedQuery = this.baseQuery?.trim();
+    if (!trimmedQuery) return;
+    if (!this.searchService) return;
 
     const params: SearchParams = {
-      query: this.filteredQuery,
+      query: trimmedQuery,
       rows: 0,
       filters: this.filterMap,
       // Fetch a few extra buckets beyond the 6 we show, in case some get suppressed
@@ -1176,7 +1178,7 @@ export class CollectionBrowser
     };
 
     this.facetsLoading = true;
-    const searchResponse = await this.searchService?.search(
+    const searchResponse = await this.searchService.search(
       params,
       this.searchType
     );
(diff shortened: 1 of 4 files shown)


src/collection-browser.ts
```diff
@@ -1498,7 +1513,9 @@ export class CollectionBrowser
    * Call this whenever the counts are invalidated (e.g., by a query change).
    */
   private refreshLetterCounts(): void {
-    this.prefixFilterCountMap = {};
+    if (Object.keys(this.prefixFilterCountMap).length > 0) {
+      this.prefixFilterCountMap = {};
+    }
```
latonv, line 1518: Just preventing needless component update cycles, by ensuring we only replace this reactive property if it wasn't already empty.

src/collection-browser.ts
```diff
@@ -541,26 +541,18 @@ export class CollectionBrowser
     });
   }
 
-  private selectedCreatorLetterChanged(): void {
-    this.creatorQuery = this.selectedCreatorFilter
-      ? `firstCreator:${this.selectedCreatorFilter}`
-      : undefined;
-  }
-
   private titleLetterSelected(
```
nsharma123, line 544: JSDocs for newly added function would great.

src/collection-browser.ts
```diff
@@ -1175,7 +1168,7 @@ export class CollectionBrowser
     };
 
     this.facetsLoading = true;
-    const searchResponse = await this.searchService?.search(
+    const searchResponse = await this.searchService.search(
```
nsharma123, line 1171: any reason why we removed optional chaining (?.) operator here?

src/collection-browser.ts
dualcnhq: +1

src/collection-browser.ts
latonv: Good question. New line 1157 above is doing an early return if there's no search service set, so the optional chain here is no longer needed.

(Same reasoning below in `fetchPage` too).

src/collection-browser.ts
latonv: Added docs for the new getters and a few other methods that were missing them before. Thanks!

PR #232, opened 2023-02-14
WEBDEV-5795 Improve accessibility for screen-readers

latonv:
This PR addresses a number of accessibility issues afflicting the collection-browser component and its subcomponents. Among the most important changes are:
- Ensuring a consistent heading hierarchy & landmarks throughout subcomponents for ease of navigability
- Adding accessible labels to some buttons/images that lacked them
- Ensuring the grid tile links are properly labeled by their item titles instead of a concatenation of all the tile text
- Upgrading JS-managed divs to native HTML widgets in some cases (e.g., `<details>`/`<summary>` for accordion widgets)

This PR also upgrades the following dependencies to versions with similar accessibility fixes included:
- `iaux-histogram-date-range`
- `iaux-infinite-scroller`


src/collection-browser.ts
```diff
@@ -1981,5 +2000,19 @@ export class CollectionBrowser
     infinite-scroller.hidden {
       display: none;
     }
+
+    .sr-only {
```
dualcnhq, line 2004: just noticed that we have a lot of `.sr-only` classes on most of the components, would it be possible to have a base `.sr-only` CSS template and reuse it by component if needed?

src/sort-filter-bar/sort-filter-bar.ts
```diff
@@ -257,12 +257,18 @@ export class SortFilterBar
...
+    const oppositeSortDirectionLabel =
+      this.sortDirection === 'asc' ? 'descending' : 'ascending';
+
     return html`
       <button
         class="sort-direction-selector"
         ?disabled=${this.selectedSort === 'relevance'}
         @click=${this.toggleSortDirection}
       >
+        <span class="sr-only"
```
dualcnhq, line 269: weird formatting of closing tags

test/sort-filter-bar/alpha-bar.test.ts
```diff
@@ -14,30 +14,34 @@ describe('Alphabetical Filter Bar', () => {
...
     el.letterCounts = { U: 10, X: 10 };
     await el.updateComplete;
 
-    // Should have exactly two letter links
-    const letterLinks = el.shadowRoot?.querySelectorAll('li > a[href]');
-    expect(letterLinks?.length).to.equal(2);
-    expect(letterLinks?.item(0).textContent?.trim()).to.equal('U');
-    expect(letterLinks?.item(1).textContent?.trim()).to.equal('X');
+    // Should have exactly two letter buttons
+    const letterButtons = el.shadowRoot?.querySelectorAll(
```
dualcnhq, line 24: :nice:

yarn.lock
```diff
@@ -108,10 +108,10 @@
   resolved "https://registry.npmjs.org/@internetarchive/field-parsers/-/field-parsers-0.1.3.tgz"
   integrity sha512-gN7ivwHWlf8V7fD3VakQ4NtddSou5i3A3TqbIo8EKF+XTxJXaXpNBmzkZEGf2oMVK8rHvrabFxZKVEy9qG7Omw==
 
-"@internetarchive/histogram-date-range@^0.1.8":
-  version "0.1.9"
-  resolved "https://registry.npmjs.org/@internetarchive/histogram-date-range/-/histogram-date-range-0.1.9.tgz"
-  integrity sha512-jXcdh7nmVxSjcAfKrMLtpEty3BNL3Y1RLenQWg44x7EpLXtAFcwRDdnMFq5SDCZqOU+tqt9cTK6IPfsWfhNkEQ==
+"@internetarchive/histogram-date-range@^0.1.10-alpha.1":
+  version "0.1.10-alpha.1"
+  resolved "https://registry.yarnpkg.com/@internetarchive/histogram-date-range/-/histogram-date-range-0.1.10-alpha.1.tgz#13df5bc3e4a086d11d96bdbf854c8ef2c7a97d60"
```
dualcnhq, line 113: how come we have npmjs registry and yarnpkg registry here?

src/collection-browser.ts
latonv: Good idea! I'll create a shared importable template for that.

src/sort-filter-bar/sort-filter-bar.ts
latonv: Agreed... but it seems like Prettier wants to force it this way. Maybe I'll just pull the whole "Change to ..." label out into a variable so that Prettier doesn't feel compelled to multi-line it.

yarn.lock
latonv: Huh, I'm not sure why that's happening. Good eyes! I'll see if another quick `yarn install` fixes things...

src/collection-browser.ts
latonv: All the `.sr-only` rules should now be drawing from a single shared template. Unfortunately, this entailed changing the indentation of some entire css blocks, which makes for a rather messy diff in a few places.

I think going forward, we should use the more flexible `static get styles() {}` form everywhere, instead of `static styles = ...`.

yarn.lock
latonv: Had to explicitly tell yarn to use the npmjs registry, but should be fixed now. That said, it probably makes little difference which registry gets used (I think they both pull from the same place in the end anyway).

src/collection-facets.ts
```diff
@@ -519,87 +527,90 @@ export class CollectionFacets extends LitElement {
   }
 
   static get styles() {
-    return css`
-      #container.loading {
-        opacity: 0.5;
-      }
+    return [
+      srOnlyStyle,
```
dualcnhq, line 531: :nice:

dualcnhq (approved):
LGTM

PR #243, opened 2023-04-06
WEBDEV-5346 Allow date picker to be expanded into a larger modal

latonv:
Adds an "Expand" button beside the date picker in the facets sidebar, which opens a modal dialog with a much larger version of the date picker that is (hopefully) easier to adjust. This larger date picker can be adjusted as needed without any immediate impact on the page until the modal's "Apply date range" button is clicked.


src/expanded-date-picker.ts
```diff
@@ -0,0 +1,144 @@
...
+   */
+  private removeEscapeListener() {
+    document.removeEventListener('keydown', this.boundEscapeListener);
+  }
+
+  /**
+   * Closes the modal dialog if the given event is pressing the Esc key.
+   * Arrow function to ensure `this` remains bound to the current component.
+   */
+  private boundEscapeListener = (e: KeyboardEvent) => {
```
iisa, line 76: this would be cool if it was also a default modal behavior...

iisa (approved):

PR #246, opened 2023-04-13
WEBDEV-5981 Make date picker full-width in mobile layout

latonv:
At mobile widths, the date picker only takes up a fraction of the available space, making it small and hard to read.

While in desktop view we solved this by allowing it to expand into a larger modal, in the mobile layout it is possible to make the date picker wider all the time, no modal required. This PR allows it to fill the entire container width available to it in mobile view.

This also entails upgrading `histogram-date-range` component version, as the previous version has a bug causing it not to fully render updates to the component's `width`/`height` properties.

### DIFF
--- a/src/collection-browser.ts
+++ b/src/collection-browser.ts
@@ -384,44 +384,75 @@ export class CollectionBrowser
   }
 
   private get collectionBrowserTemplate() {
-    const shouldShowSearching =
-      this.searchResultsLoading || this.totalResults === undefined;
-    const resultsCount = this.totalResults?.toLocaleString();
-    const resultsLabel = this.totalResults === 1 ? 'Result' : 'Results';
-    return html` <div id="left-column-scroll-sentinel"></div>
+    return html`
+      <div id="left-column-scroll-sentinel"></div>
+      ${this.leftColumnTemplate} ${this.rightColumnTemplate}
+    `;
+  }
+
+  private get leftColumnTemplate(): TemplateResult {
+    if (this.mobileView) {
+      return this.mobileLeftColumnTemplate;
+    }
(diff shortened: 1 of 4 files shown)


src/collection-browser.ts
```diff
@@ -384,44 +386,75 @@ export class CollectionBrowser
   }
 
   private get collectionBrowserTemplate() {
```
latonv, line 388: The old version of this template was getting a bit bloated and hard to read, so I've split it out into several smaller sub-templates that should hopefully be easier to understand.

src/collection-browser.ts
```diff
@@ -869,7 +903,8 @@ export class CollectionBrowser
   handleResize(entry: ResizeObserverEntry): void {
     const previousView = this.mobileView;
     if (entry.target === this.contentContainer) {
-      this.mobileView = entry.contentRect.width < this.mobileBreakpoint;
+      this.contentWidth = entry.contentRect.width;
+      this.mobileView = this.contentWidth < this.mobileBreakpoint;
```
latonv, line 907: Preserving the contentRect's width to be passed down into the date picker (which requires an explicit width to render its svg).

src/collection-facets.ts
```diff
@@ -229,10 +231,6 @@ export class CollectionFacets extends LitElement {
       : nothing;
   }
 
-  private get currentYearsHistogramAggregation(): Aggregation | undefined {
-    return this.aggregations?.year_histogram;
-  }
-
```
latonv, line 235: (Obsolete, unused method)

iisa (approved):
looks good! thanks for the walk thru and splitting up code for good practice ^_^

PR #247, opened 2023-04-24
Sort bar options should stay put when selected

nsharma123:
https://webarchive.jira.com/browse/WEBDEV-5782

### DIFF
--- a/src/sort-filter-bar/sort-filter-bar.ts
+++ b/src/sort-filter-bar/sort-filter-bar.ts
@@ -397,7 +397,7 @@ export class SortFilterBar
     return html`
       <button
         class=${isSelected ? 'selected' : nothing}
-        title="${displayName}"
+        data-title="${displayName}"
         @click=${(e: Event) => {
           e.preventDefault();
           options?.onClick?.(e);
@@ -448,7 +448,7 @@ export class SortFilterBar
         <span
           class="dropdown-label"
           slot="dropdown-label"
-          title="${options.displayName.values}"
+          data-title="${options.displayName.values}"
           @click=${options.onLabelInteraction ?? nothing}
           @keydown=${options.onLabelInteraction
             ? (e: KeyboardEvent) => {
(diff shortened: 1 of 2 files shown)


src/sort-filter-bar/sort-filter-bar.ts
```diff
@@ -959,7 +959,7 @@ export class SortFilterBar
         }
 
         #desktop-sort-selector li button.selected {
-          font-weight: bold;
+          text-shadow: rgb(0 0 0) 0.08rem 0 0;
```
latonv, line 962: This _does_ address the text shifting problem, but I find the text shadows look noticeably different from the actual boldface font. I'm not sure inconsistent font appearances are a good tradeoff for fixing the slight shifts.

One alternative might be to place an always-bolded `visibility: hidden` element with the same content underneath each sort button, essentially reserving the needed space invisibly. Might take a bit of fiddling to get the positioning right, but that way we could keep the desired bold fonts. What do you think?

nsharma123: Hi Laton!
Thanks for raising the question around font appearances. I applied another approach to keep using bold but handling the text shift problem is gone. Please go through the code changes and again do QA. Thanks!!

src/sort-filter-bar/sort-filter-bar.ts
```diff
@@ -962,6 +964,19 @@ export class SortFilterBar
...
+        #desktop-sort-selector li button::before,
+        #desktop-sort-selector .dropdown-label::before {
+          display: block;
+          content: attr(title);
+          font-weight: bold;
+          height: 0;
+          overflow: hidden;
+          visibility: hidden;
+        }
+
```
latonv, line 979: Nice elegant approach!

Minor suggestion: could we use an attribute other than `title` for this, so that we don't get tooltips when hovering the labels? Something like `data-content`?

latonv (approved):
LGTM! Just one very minor suggestion about which attribute should hold the sizing content.

PR #250, opened 2023-04-27
WEBDEV-5813 Mediatype icon links should open in the same tab

latonv:
Currently, clicking the mediatype icon on a list view tile opens that mediatype's collection in a new tab. This PR just makes it open in the same tab for consistency with all the other links on the page.


iisa (approved):
🚢

PR #251, opened 2023-04-27
WEBDEV-5943 Use meaningful hashes for the `uid` param on search requests

latonv:
Currently the `uid` param that gets passed to the PPS on searches consists of the full query w/ all filters and sort options verbatim. That has made it a challenge to debug issues with search requests on the backend, so this PR produces a more easily traceable `uid` structured from short hashes and indicating the page requested, kind of request (hits, aggregations, or both), and the timestamp.


iisa (approved):
ok with me pending passing tests 👍

PR #253, opened 2023-05-16
WEBDEV-6098 Update date picker expand icon & add analytics

latonv:
This PR replaces the old icon for expanding the date picker with one that is less visually prominent. It also sends analytics events when the date picker is expanded and when a date range is applied via the modal.


src/collection-facets.ts
```diff
@@ -191,16 +193,25 @@ export class CollectionFacets extends LitElement {
...
+      category: analyticsCategories.default,
+      action: analyticsActions.histogramExpanded,
+      label: 'expanded',
     });
   }
 
+  private handleExpandedDatePickerClosed = (): void => {
+    this.modalManager?.classList.remove('expanded-date-picker');
+  };
+
```
latonv, line 214: (A bit of refactoring so that `<expanded-date-picker>` doesn't need require any knowledge of the corresponding modal class.)

src/collection-facets.ts
```diff
@@ -191,16 +193,25 @@ export class CollectionFacets extends LitElement {
...
-      userClosedModalCallback: () => {
-        this.modalManager?.classList.remove('expanded-date-picker');
-      },
+      userClosedModalCallback: this.handleExpandedDatePickerClosed,
+    });
+
+    this.analyticsHandler?.sendEvent({
+      category: analyticsCategories.default,
+      action: analyticsActions.histogramExpanded,
+      label: 'expanded',
```
bfalling, line 207: QUESTION: Often, we put the page's URL in the `label`. Would it be more useful to put the URL or the collection identifier here?

src/expanded-date-picker.ts
```diff
@@ -105,15 +114,28 @@ export class ExpandedDatePicker extends LitElement {
...
+
+    const yearQuery =
+      this.maxSelectedDate && this.minSelectedDate
+        ? `year:[${this.minSelectedDate} TO ${this.maxSelectedDate}]`
+        : undefined;
+
+    this.analyticsHandler?.sendEvent({
+      category: analyticsCategories.default,
+      action: analyticsActions.histogramChangedFromModal,
+      label: yearQuery,
```
bfalling, line 126: QUESTION: Often, we put the page's URL in the label. Would it be more useful to put the URL or the collection identifier here?

latonv: Yeah, I agree that's more helpful. Both events now use the URL (which will include any collection identifier anyway).

src/collection-facets.ts
latonv: Done -- label is now the URL.

PR #254, opened 2023-05-27
WEBDEV-6108 Always render `mediatype: collection` facet when present

latonv:
Ensures that when the current search results include a `mediatype: collection` aggregation bucket, its facet is rendered in the sidebar regardless of its bucket count.

### DIFF
--- a/test/collection-facets.test.ts
+++ b/test/collection-facets.test.ts
@@ -10,7 +10,11 @@ import {
 import type { CollectionFacets } from '../src/collection-facets';
 import '@internetarchive/modal-manager';
 import '../src/collection-facets';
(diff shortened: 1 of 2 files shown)


test/collection-facets.test.ts
```diff
@@ -533,6 +537,126 @@ describe('Collection Facets', () => {
...
+          },
+          {
+            key: 'collection',
+            doc_count: 1,
+          },
+        ],
+      }),
+    };
+
+    const selectedFacets = getDefaultSelectedFacets();
```
iisa, line 638: 👏

iisa (approved):

PR #255, opened 2023-05-30
WEBDEV-6122 Collapse double-wrapped quotes in tile href

latonv:
When the backend provides a tile `__href__` field for FTS queries, its values sometimes have erroneous ""doubly quoted"" clauses which cause subsequent search-within to fail. While there are more fundamental backend fixes in the works to address the source of those extra quotes, this PR provides a temporary fix to prevent them from affecting the rendered tile links.

### DIFF
--- a/src/collection-browser.ts
+++ b/src/collection-browser.ts
@@ -1786,7 +1786,7 @@ export class CollectionBrowser
         dateReviewed: result.reviewdate?.value,
         description: result.description?.values.join('\n'),
         favCount: result.num_favorites?.value ?? 0,
-        href: result.__href__?.value,
+        href: this.collapseRepeatedQuotes(result.__href__?.value),
         identifier: result.identifier,
         issue: result.issue?.value,
         itemCount: result.item_count?.value ?? 0,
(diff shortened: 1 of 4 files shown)


src/collection-browser.ts
```diff
@@ -1786,7 +1786,7 @@ export class CollectionBrowser
         dateReviewed: result.reviewdate?.value,
         description: result.description?.values.join('\n'),
         favCount: result.num_favorites?.value ?? 0,
-        href: result.__href__?.value,
+        href: this.collapseRepeatedQuotes(result.__href__?.value),
```
iisa, line 1789: q: does this need the petabox fix to go with it?

src/collection-browser.ts
```diff
@@ -1811,6 +1811,20 @@ export class CollectionBrowser
...
+   * ""clauses wrapped in two sets of quotes"". This assumes the quotes are already
+   * URL-encoded.
+   *
+   * This should be a temporary measure to address the fact that the __href__ field
+   * sometimes acquires extra quotation marks during query rewriting. Once there is a
+   * full Lucene parser in place that handles quoted queries correctly, this can likely
+   * be removed.
+   */
+  private collapseRepeatedQuotes(str?: string): string | undefined {
+    return str?.replace(/%22%22(?!%22%22)(.+?)%22%22/g, '%22$1%22');
```
iisa, line 1825: q: if it is undefined, will the display bork?

iisa (approved):
looks good, had question around double quote check returning undefined.
also, does this need the service update to work?

src/collection-browser.ts
latonv: No, the two fixes should be independent. This one only impacts the search page, while the petabox fix only impacts the (non-Offshoot) collection pages.

src/collection-browser.ts
latonv: The case where `__href__` is undefined should maintain the same behavior as before: leaving it undefined in the tile model, and allow the tile components to decide how to render the tile link.

So if `collapseRepeatedQuotes` receives an undefined input, it's expected that it should output undefined to match that prior behavior.

PR #256, opened 2023-05-31
WEBDEV-5910 Accept `withinCollection` prop with no query, for collection pages

latonv:
To enable use on collection pages, the collection browser must be able to display the contents of a collection without any additional query set. The search service already allows this via `pageType=collection_details` with a valid `pageTarget` set to the collection identifier.

This PR adds the `withinCollection` prop to accept the collection identifier, passing it to the search service with the correct page type, and functioning correctly with an empty query in the presence of this additional prop.

### DIFF
--- a/src/app-root.ts
+++ b/src/app-root.ts
@@ -95,9 +95,7 @@ export class AppRoot extends LitElement {
     this.searchQuery = this.baseQueryField.value;
     this.collectionBrowser.searchType = this.searchType;
 
-    if ((this.currentPage ?? 1) > 1) {
-      this.collectionBrowser.goToPage(this.currentPage ?? 1);
-    }
+    this.goToCurrentPage();
   }
 
   private collectionChanged(e: Event) {
@@ -105,8 +103,13 @@ export class AppRoot extends LitElement {
     this.withinCollection = this.baseCollectionField.value;
     this.collectionBrowser.withinCollection = this.withinCollection;
 
-    if ((this.currentPage ?? 1) > 1) {
-      this.collectionBrowser.goToPage(this.currentPage ?? 1);
+    this.goToCurrentPage();
+  }
+
+  private goToCurrentPage() {
+    const page = this.currentPage ?? 1;
+    if (page > 1) {
+      this.collectionBrowser.goToPage(page);
     }
   }
 
--- a/src/app-root.ts
+++ b/src/app-root.ts
@@ -39,6 +39,8 @@ export class AppRoot extends LitElement {
 
   @state() private searchQuery?: string;
 
+  @state() private withinCollection?: string;
+
   @state() private cellWidth: number = 18;
 
   @state() private cellHeight: number = 29;
@@ -54,6 +56,9 @@ export class AppRoot extends LitElement {
   @property({ type: Object, reflect: false }) latestAction?: AnalyticsEvent;
 
   @query('#base-query-field') private baseQueryField!: HTMLInputElement;
+
+  @query('#base-collection-field')
+  private baseCollectionField!: HTMLInputElement;
 
   @query('#page-number-input') private pageNumberInput!: HTMLInputElement;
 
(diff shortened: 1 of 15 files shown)


src/collection-browser.ts
```diff
@@ -118,6 +120,8 @@ export class CollectionBrowser
 
   @property({ type: Boolean }) showHistogramDatePicker = false;
 
+  @property({ type: String }) collectionPagePath: string = '/details/';
```
latonv, line 123: By default, links to other collections point at their /details/<id> page. This enables customizing that behavior, e.g. to make them point at /collection/<id> instead.

src/collection-browser.ts
```diff
@@ -1702,10 +1724,12 @@ export class CollectionBrowser
...
     // item count to the real total number of results (rather than the
     // temporary estimates based on pages rendered so far).
-    if (results.length < this.pageSize) {
+    const resultCountDiscrepancy = numRows - results.length;
+    if (resultCountDiscrepancy > 0) {
       this.endOfDataReached = true;
       if (this.infiniteScroller) {
-        this.infiniteScroller.itemCount = this.totalResults;
+        this.infiniteScroller.itemCount =
+          this.estimatedTileCount - resultCountDiscrepancy;
```
latonv, line 1732: This is to work around some `filter-top-level` issues on collections. Because the `totalResults` value sent by the PPS doesn't always match the actual number of results it sends, we can't rely on it to cap off the infinite scroller's item count.

src/app-root.ts
```diff
@@ -95,6 +100,16 @@ export class AppRoot extends LitElement {
     }
   }
 
+  private collectionChanged(e: Event) {
+    e.preventDefault();
+    this.withinCollection = this.baseCollectionField.value;
+    this.collectionBrowser.withinCollection = this.withinCollection;
+
+    if ((this.currentPage ?? 1) > 1) {
+      this.collectionBrowser.goToPage(this.currentPage ?? 1);
```
bfalling, line 109: SUGGESTION: Use meaningful temp variable instead of calc'ing `this.currentPage ?? 1` twice?

src/tiles/list/tile-list-compact.ts
```diff
@@ -66,9 +68,14 @@ export class TileListCompact extends LitElement {
...
-      ? `${this.baseNavigationUrl}${this.model.href}`
-      : `${this.baseNavigationUrl}/details/${this.model?.identifier}`;
+    if (this.model?.href) {
+      return `${this.baseNavigationUrl}${this.model?.href}`;
+    }
+
+    const isCollection = this.model?.mediatype === 'collection';
+    return `${this.baseNavigationUrl}${
+      isCollection ? this.collectionPagePath : '/details/'
+    }${this.model?.identifier}`;
```
bfalling, line 78: QUESTION: Does this optional chain value need a protection against being undefined? e.g. `?? ''`?

src/tiles/tile-dispatcher.ts
```diff
@@ -140,9 +142,14 @@ export class TileDispatcher
   private get linkTileHref() {
     // Use the server-specified href if available.
     // Otherwise, construct a details page URL from the item identifier.
-    return this.model?.href
-      ? `${this.baseNavigationUrl}${this.model?.href}`
-      : `${this.baseNavigationUrl}/details/${this.model?.identifier}`;
+    if (this.model?.href) {
+      return `${this.baseNavigationUrl}${this.model?.href}`;
```
bfalling, line 146: Same question about whether necessary to trap for `undefined`.

src/tiles/tile-dispatcher.ts
```diff
@@ -140,9 +142,14 @@ export class TileDispatcher
...
+    }
+
+    const isCollection = this.model?.mediatype === 'collection';
+    return `${this.baseNavigationUrl}${
+      isCollection ? this.collectionPagePath : '/details/'
+    }${this.model?.identifier}`;
```
bfalling, line 152: Ditto.

bfalling:
Some minor questions, 1 suggestion. Otherwise, LGTM!

src/tiles/list/tile-list-compact.ts
latonv: Yes -- in fact we can probably just have an early-out at the top of the method to ensure the model is defined. If it's not, there's little point in trying to render the link anyway.

PR #261, opened 2023-06-21
WEBDEV-6167 Make tiles show creator matching letter filter, if one is applied

latonv:
Currently, search results in grid view & compact list view only show the first creator when there are multiple creators for the item. However, the “initial letter” filters (available when sorting by creator) match _any_ of the creators in the list, not just the first. So it can look strange when such a filter is applied, but results still show a creator that doesn’t match the filter.

This PR makes search results with multiple creators adjust to show the first _matching_ creator on the tile, rather than the first overall.


src/tiles/grid/item-tile.ts
```diff
@@ -78,12 +80,24 @@ export class ItemTile extends LitElement {
...
-    if (!this.model?.creator) return nothing;
+    let displayedCreator = this.model?.creator;
+
+    // If we're filtering by creator initial and have multiple creators, we want
+    // to surface the first creator who matches the filter.
+    if (this.selectedCreatorFilter && this.model?.creators.length) {
+      const firstLetter = this.selectedCreatorFilter;
+      displayedCreator =
+        this.model.creators.find(creator =>
+          creator.toUpperCase().startsWith(firstLetter)
```
cdrini, line 91: Chat GPT recommends this to make it so that e.g. "E" still matches "émile" (which the search engine will match).

```suggestion
          creator..normalize("NFD").replace(/[\u0300-\u036f]/g, "").toUpperCase().startsWith(firstLetter)
```

src/tiles/grid/item-tile.ts
```diff
@@ -78,12 +80,24 @@ export class ItemTile extends LitElement {
    * Templates
    */
   private get creatorTemplate(): TemplateResult | typeof nothing {
-    if (!this.model?.creator) return nothing;
+    let displayedCreator = this.model?.creator;
+
+    // If we're filtering by creator initial and have multiple creators, we want
+    // to surface the first creator who matches the filter.
+    if (this.selectedCreatorFilter && this.model?.creators.length) {
+      const firstLetter = this.selectedCreatorFilter;
```
cdrini, line 88: I'd drop the `firstLetter` declaration ; `this.selectedCreatorFilter` is pretty clear.

src/tiles/list/tile-list-compact.ts
```diff
@@ -80,6 +82,22 @@ export class TileListCompact extends LitElement {
     return `${this.baseNavigationUrl}${basePath}${this.model.identifier}`;
   }
 
+  private get creator(): string | typeof nothing {
```
cdrini, line 85: This is a little duplicate ? Maybe putting a helper function on `model` might be cleaner, cause then you can do `this.model?.getDisplayedCreator(selectedCreatorFilter)`.

src/tiles/list/tile-list-compact.ts
```diff
@@ -24,6 +24,8 @@ export class TileListCompact extends LitElement {
 
   @property({ type: Object }) sortParam: SortParam | null = null;
 
+  @property({ type: String }) selectedCreatorFilter?: string;
```
cdrini, line 27: Selected feels a bit redundant; `createrFilter` or `creatorLetterFilter` seem a little clearer to me.

```suggestion
  @property({ type: String }) creatorFilter?: string;
```

cdrini (approved):
Lgtm! A few code readability/thoughts, but everything is just suggestions. Seems solid!

src/tiles/grid/item-tile.ts
latonv: I agree in principle, but TypeScript can't guarantee `this.selectedCreatorFilter` isn't undefined within the `find` callback (_technically_ correct, I suppose). So it was either this or an explicit type assertion of some sort.

cdrini: ahhhh gotcha didn't think of that! Makes sense

src/tiles/list/tile-list-compact.ts
latonv: Yeah, I'll try rethinking this a bit. It's tough because two of our three display modes use this new handling, while the third displays things another way entirely. I'm not sure I want the tile model to have to know about the display logic, but maybe this can be abstracted out somewhere cleaner.

src/tiles/grid/item-tile.ts
latonv: Good catch! I wonder if the `replace` step is actually necessary here... If that range of code points only includes combining characters that modify the character _before_ them, then the decomposition alone should still start with the bare letter... I think? (I'll subject this to some more testing 😄)

cdrini: Oh good thinking! ChatGPT agrees 😁 https://chat.openai.com/share/0d1e3d87-9515-40fb-a010-23f3d8bdb11d

PR #265, opened 2023-07-05
WEBDEV-6176 Set default sort option from collection metadata

latonv:
On collection pages, there may be a default sort applied to the results independent of the user's sort selection. For instance, most collections have a default sort of `week:desc` (this can be modified in collection metadata).

Accordingly, when browsing collection members, this default sort should be represented accurately in the sort bar UI. This PR slightly refactors how sort fields are handled, adding a `default` SortField enum member which indicates deferral to the current defaults (`relevance` if viewing a search results page, or the current collection's default if viewing a collection page).

### DIFF
--- a/src/collection-browser.ts
+++ b/src/collection-browser.ts
@@ -1780,7 +1780,7 @@ export class CollectionBrowser
 
     if (this.withinCollection) {
       this.collectionInfo = success.response.collectionExtraInfo;
-      
+
       // For collections, we want the UI to respect the default sort option
       // which can be specified in metadata, or otherwise assumed to be week:desc
       this.applyDefaultCollectionSort(success.response.collectionExtraInfo);
--- a/src/collection-browser.ts
+++ b/src/collection-browser.ts
@@ -564,6 +564,7 @@ export class CollectionBrowser
         .defaultSortDirection=${this.defaultSortDirection}
         .selectedSort=${this.selectedSort}
         .sortDirection=${this.sortDirection}
+        .showRelevance=${this.isRelevanceSortAvailable}
         .displayMode=${this.displayMode}
         .selectedTitleFilter=${this.selectedTitleFilter}
         .selectedCreatorFilter=${this.selectedCreatorFilter}
@@ -1204,6 +1205,11 @@ export class CollectionBrowser
       this.infiniteScroller.reload();
     }
 
+    if (this.withinCollection && this.baseQuery?.trim()) {
+      this.defaultSortField = SortField.relevance;
+      this.defaultSortDirection = null;
+    }
+
     if (!this.initialQueryChangeHappened && this.initialPageNumber > 1) {
       this.scrollToPage(this.initialPageNumber);
     }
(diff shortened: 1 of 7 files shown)


test/mocks/mock-search-responses.ts
```diff
@@ -512,6 +512,86 @@ export const getMockSuccessExtraQuotedHref: () => Result<
...
 });
 
+export const getMockSuccessWithDefaultSort: () => Result<
+  SearchResponse,
+  SearchServiceError
+> = () => ({
+  success: {
+    request: {
+      clientParameters: {
+        user_query: 'default-sort',
```
iisa, line 522: 👍

test/collection-browser.test.ts
```diff
@@ -988,6 +989,82 @@ describe('Collection Browser', () => {
...
+    await el.updateComplete;
+    await aTimeout(50);
+
+    const sortBar = el.shadowRoot?.querySelector(
+      'sort-filter-bar'
+    ) as SortFilterBar;
+    expect(sortBar).to.exist;
+    expect(sortBar.defaultSortField).to.equal(SortField.dateadded);
+    expect(sortBar.defaultSortDirection).to.equal('desc');
+    expect(sortBar.selectedSort).to.equal(SortField.default);
```
iisa, line 1038: 👍

src/models.ts
```diff
@@ -99,9 +101,10 @@ export const SortFieldDisplayName: {
 };
 
 export const DefaultSortDirection: {
-  [key in SortField]: SortDirection;
+  [key in SortField]: SortDirection | null;
 } = {
-  relevance: 'desc', // Can't actually change the sort direction for relevance
+  default: null,
+  relevance: null, // Can't actually change the sort direction for relevance
```
iisa, line 107: q: if this changed to null, does that imply that it can change? the comment says it cannot change value?

src/collection-browser.ts
```diff
@@ -1797,6 +1827,39 @@ export class CollectionBrowser
     this.collectionNameCache?.preloadIdentifiers(collectionIdsArray);
   }
 
+  /**
+   * Applies any default sort option for the current collection, by checking
+   * for one in the collection's metadata. If none is found, defaults to sorting
+   * descending by weekly views.
+   */
+  private applyDefaultCollectionSort(collectionInfo?: CollectionExtraInfo) {
```
iisa, line 1835: 👍

iisa (approved):
:lgtm:

src/models.ts
latonv: Oops, sounds like my comment wasn't clear enough! It's supposed to allude to how we disable the sort direction button when Relevance sort is selected -- i.e., the direction is not _user_-changeable in that case. So the `'desc'` value didn't actually do anything here, a potential source of confusion.

I can update to word this better. And FWIW, I have another in-flight PR that gathers up these various sort-related objects into a single declarative config which should hopefully be simpler to work with.

PR #268, opened 2023-07-13
WEBDEV-6209 Fix More facets within a collection

latonv:
Currently, the More facet links are broken when searching within a collection (the dialog just hangs). This PR ensures the More facets search query uses the correct `page_type` and `page_target` params, fixing the issue.


nsharma123 (approved):
Looks good to me!

PR #269, opened 2023-07-13
WEBDEV-6171 Handle unrecognized sort options

latonv:
Currently, one cannot set a sort option via URL if it does not appear in the sort bar -- collection browser clobbers these. So valid sorts like `avg_rating` are inaccessible. However, these were allowed on the legacy search page, and consequently there exist old links and possibly old bookmarks that include these now-unrecognized sort params.

To ensure these sort options continue to be available (even if not surfaced in the UI), this PR allows unrecognized sort options to be specified in the URL and passed as-is to the backend, which already has handling to determine whether they are valid.

Moreover, this PR includes a refactoring pass to clean up how various properties of sort options are handled internally, centralizing them into a single config instead of having several individual maps for different properties.

### DIFF
--- a/src/models.ts
+++ b/src/models.ts
@@ -56,6 +56,7 @@ export type CollectionBrowserContext = 'collection' | 'search';
  */
 export enum SortField {
   'default' = 'default',
+  'unrecognized' = 'unrecognized',
   'relevance' = 'relevance',
   'alltimeview' = 'alltimeview',
   'weeklyview' = 'weeklyview',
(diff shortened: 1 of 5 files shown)


src/models.ts
```diff
@@ -67,120 +68,202 @@ export enum SortField {
...
-} = {
-  week: 'week',
-  downloads: 'downloads',
-  titleSorter: 'title',
-  date: 'date',
-  publicdate: 'publicdate',
-  reviewdate: 'reviewdate',
-  addeddate: 'addeddate',
-  creatorSorter: 'creator',
+export const SORT_OPTIONS: Record<SortField, SortOption> = {
```
latonv, line 123: This map consolidates all the properties of each sort field into a single config object, so that we don't have to hunt them down across a bunch of other maps. This should hopefully be easier & more organized than before.

In particular:
- the previous `SortFieldDisplayName` map became the `displayName` property.
- the previous `DefaultSortDirection` map became the `defaultSortDirection` property.
- the previous `MetadataFieldToSortField`, `MetadataFieldToURLField`, and `URLFieldToSortField` maps are now all handled by the `urlNames` property, which serves to list all the valid aliases for a given sort option that are recognized from the URL or backend APIs.
- the previous `SortFieldToMetadataField` map is now handled by the `searchServiceKey` property, indicating the correct key to request that sort from the search service.
- we add the `canSetDirection`, `shownInSortBar`, `shownInURL`, and `handledBySearchService` properties as a more declarative alternative to some of the previous sort bar & URL logic that addressed these as special cases.

src/models.ts
```diff
@@ -67,120 +68,202 @@ export enum SortField {
...
-  creator: SortField.creator,
-  creatorSorter: SortField.creator,
-};
+export function sortOptionFromAPIString(sortName?: string | null): SortOption {
+  return (
+    Object.values(SORT_OPTIONS).find(opt =>
+      opt.urlNames.some(name => sortName === name)
+    ) ?? SORT_OPTIONS[SortField.unrecognized]
+  );
+}
```
latonv, line 266: Helper function to look up the sort option that corresponds to a key from the URL or backend API.

src/restoration-state-handler.ts
```diff
@@ -1,29 +1,22 @@
...
-  MetadataFieldToURLField,
+  sortOptionFromAPIString,
+  SORT_OPTIONS,
 } from './models';
 import { arrayEquals } from './utils/array-equals';
 
 export interface RestorationState {
   displayMode?: CollectionDisplayMode;
   searchType?: SearchType;
-  sortParam?: SortParam;
```
latonv, line 26: This property was redundant as we're already passing the sort field & direction that are used to derive it.

iisa (approved):

src/restoration-state-handler.ts
```diff
@@ -354,7 +356,31 @@ export class RestorationStateHandler
     return restorationState;
   }
 
-  // remove optional opening and closing quotes from a string
+  /**
+   * Converts a URL sort param into a field/direction pair, if possible.
+   * Either or both may be undefined if the param is not in a recognized format.
+   */
+  private getSortFieldAndDirection(sortParam: string) {
```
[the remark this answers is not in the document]
latonv: Sort of -- this parsing code already existed above, in the restoration method. I just pulled it out into its own method for DRY purposes, since the same functionality is needed in the state persistence method now too.

PR #273, opened 2023-07-26
WEBDEV-6227 Add "Part Of" section above facets

latonv:
The old collection page had a "Part of" section in the left sidebar which listed & linked to the parent collections. This PR adds that feature to the new page as well.


iisa (approved):
nice tests!

PR #275, opened 2023-07-30
WEBDEV-6259 Ensure all URL params are respected on initial collection page load

latonv:
A bug introduced in #272 causes the collection page to clobber URL params such as the sort & facets when visiting a collection page that has a default sort. The cause of this was that a line which clears filters when the target collection is changed, which was not properly guarded to prevent it from running on the initial page load (i.e., when the target collection is changed from `undefined` to its initial value).

This PR addresses that bug and adds a suite of several tests designed to avert any further issues of this nature.

### DIFF
--- a/test/collection-browser.test.ts
+++ b/test/collection-browser.test.ts
@@ -1246,4 +1246,180 @@ describe('Collection Browser', () => {
     // property is reflected as attribute
     expect(el.getAttribute('searchcontext')).to.equal('unicorn-search');
   });
+
+  it('respects the initial set of URL parameters for a general search', async () => {
+    const url = new URL(window.location.href);
+    const { searchParams } = url;
+    searchParams.set('query', 'foo');
+    searchParams.set('sin', 'TXT');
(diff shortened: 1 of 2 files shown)


test/collection-browser.test.ts
```diff
@@ -1246,4 +1246,180 @@ describe('Collection Browser', () => {
...
+    expect(el.withinCollection).to.equal('bar');
+    expect(el.baseQuery).to.equal('foo');
+    expect(el.searchType).to.equal(SearchType.FULLTEXT);
+    expect(el.selectedSort).to.equal(SortField.default);
+    expect(el.selectedFacets?.mediatype?.data).not.to.exist;
+    expect(el.selectedFacets?.subject?.baz).not.to.exist;
+    expect(el.selectedTitleFilter).not.to.exist;
+    expect(el.minSelectedDate).not.to.exist;
+    expect(el.maxSelectedDate).not.to.exist;
+  });
```
nsharma123, line 1424: Nice set of tests! LGTM!

PR #278, opened 2023-08-01
WEBDEV-6246 Add management UI to remove items

latonv:
This PR ports over functionality from the legacy pages to remove items from a collection. A manage bar is added which replaces the sort bar UI when manage mode is active, and checkboxes are displayed on tiles when in this state, allowing a batch of them to be selected and removed all at once.


dualcnhq (approved):
LGTM

PR #279, opened 2023-08-04
WEBDEV-6256 Show more appropriate error message when collection fails to load

latonv:
Currently when the PPS request for a collection encounters an error, we are at best displaying an error message that implies something is wrong with the query (even when there isn't one), or at worst showing the "collection not found" state.

This PR adds a new placeholder option that should be somewhat more user-friendly, noting that the search engine couldn't load the collection details and linking to our support channel in case the problem persists.

### DIFF
--- a/src/empty-placeholder.ts
+++ b/src/empty-placeholder.ts
@@ -48,9 +48,9 @@ export class EmptyPlaceholder extends LitElement {
       Tips for constructing search queries.
     </a> `);
 
-  private static readonly MESSAGE_COLLECTION_ERROR = msg(html` The search engine
-    encountered an error fetching details for this collection. If the problem
-    persists, please let us know at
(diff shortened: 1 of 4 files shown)


src/empty-placeholder.ts
```diff
@@ -113,6 +120,16 @@ export class EmptyPlaceholder extends LitElement {
     `;
   }
 
+  private get collectionErrorTemplate(): TemplateResult {
```
iisa, line 123: q: does this fire a sentry error or anything when displayed? should we?

iisa (approved):
looks good, had 1 question

src/empty-placeholder.ts
latonv: We fire a Sentry message whenever we get an error back from the PPS, so this is already covered.

PR #282, opened 2023-08-08
Dispatch event to switch on ?tab=about tab when resultCount is 0

nsharma123:
dispatch an event to collection-page if resultcount is zero. The event will be listened at offshoot and switch to ?tab=about to show about page of collection.

### DIFF
--- a/package.json
+++ b/package.json
@@ -3,7 +3,7 @@
   "description": "The Internet Archive Collection Browser.",
   "license": "AGPL-3.0-only",
   "author": "Internet Archive",
-  "version": "1.13.0-alpha2",
+  "version": "1.13.0-alpha3",
   "main": "dist/index.js",
   "module": "dist/index.js",
   "scripts": {
--- a/package.json
+++ b/package.json
@@ -3,7 +3,7 @@
   "description": "The Internet Archive Collection Browser.",
   "license": "AGPL-3.0-only",
   "author": "Internet Archive",
-  "version": "1.13.0",
+  "version": "1.13.0-alpha2",
   "main": "dist/index.js",
   "module": "dist/index.js",
   "scripts": {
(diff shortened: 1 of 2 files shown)


src/collection-browser.ts
```diff
@@ -1125,6 +1125,14 @@ export class CollectionBrowser
...
   }
 
+  private emitTabChanged(tabId: string) {
+    this.dispatchEvent(
+      new CustomEvent<String>('tabChanged', {
+        detail: tabId,
+      })
+    );
+  }
+
```
latonv, line 1135: Suggestion: Let's make it so collection browser doesn't need to have knowledge of its parent component's tabs. Maybe we just emit an `emptyResults` event with no args, and let the parent handle that as appropriate?

nsharma123: I used custom event with args to make is reusable if future if we want to communicate with offfshoot later to switch between tabs. that way, we can simply send tabId to collection-browser to switch.

iisa: +1 to emptyResults bc tab is not direct child of collection browser, it's slotted 👍 please do update to emptyResults

iisa (changes requested):

package.json
```diff
@@ -3,7 +3,7 @@
   "description": "The Internet Archive Collection Browser.",
   "license": "AGPL-3.0-only",
   "author": "Internet Archive",
-  "version": "1.13.0",
+  "version": "1.13.0-alpha3",
```
latonv, line 6: Just a reminder to revert this version change before merging.

latonv (approved):
Thanks for updating the event, LGTM!

nsharma123 (commented):
Thanks. Changed.

PR #290, opened 2023-08-28
Test ia-dropdown component change

mc2:
Verify new ia-dropdown change makes no change in sort bar.

WEBDEV-6353


mc2:
tested and closed

PR #291, opened 2023-08-30
Collection page tiles should still show date line when default sorting by a date

dualcnhq:
### Issue:

When sorting by a type of date explicitly, the result tiles correctly show the corresponding date value in place of the creator line correctly and tiles should display the `defaultSort` if it's available within collection search

### Testing:

- load the demo app and search for `inlibrary` as collection query
- Date published is set as sortField param and the `published` text is displayed in the item-tile
- notice that the url didn't change with `?sort=` query param

Screenshot 2023-08-30 at 20 54 32

### DIFF
--- a/src/collection-browser.ts
+++ b/src/collection-browser.ts
@@ -103,6 +103,8 @@ export class CollectionBrowser
 
   @property({ type: Object }) sortParam: SortParam | null = null;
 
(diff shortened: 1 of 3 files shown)


latonv (approved):
LGTM, thanks for handling the default cases!

PR #294, opened 2023-09-11
Slot for user-facet list template

dualcnhq:


src/collection-browser.ts
```diff
@@ -493,12 +521,27 @@ export class CollectionBrowser
...
+    `;
+  }
+
   /**
    * Template for the desktop version of the left column, displaying the facets sidebar.
    */
   private get desktopLeftColumnTemplate(): TemplateResult {
     return html`
       <div id="left-column" class="column">
+        ${this.isUserFacetView ? this.userFacetListTemplate : nothing}
```
dualcnhq, line 543: <div>
<slot>just add the colored divs here for now</slot>
</div>

PR #295, opened 2023-09-12
Slot for facet-top template

dualcnhq:
## Changes:

- add a named slot into the top of the facets column
- add div container with border and colored div for now to represent slot container
- add adjustable width and height variable for div slot container
- add checkbox toggle to show/hide the facet-top view under dev-tools toggle controls

[facet-top-container.gif](https://github.com/internetarchive/iaux-collection-browser/assets/1281581/d11802ce-3f8c-459b-bd0a-8f1bb8e4c04e)

Note: 
~~no need to create new version of npm package~~
will cut a new patch tag to update in offshoot

### DIFF
--- a/src/collection-browser.ts
+++ b/src/collection-browser.ts
@@ -2390,8 +2390,6 @@ export class CollectionBrowser
 
         #facet-top-view {
           display: flex;
-          margin-top: 10px;
-          margin-bottom: 10px;
         }
 
         /**
--- a/src/collection-browser.ts
+++ b/src/collection-browser.ts
@@ -493,21 +493,15 @@ export class CollectionBrowser
     `;
   }
 
-  private get facetTopViewTemplate(): TemplateResult {
-    return html`
-      <div id="facet-top-view">
-        <slot name="facet-top"></slot>
-      </div>
-    `;
-  }
-
   /**
    * Template for the desktop version of the left column, displaying the facets sidebar.
    */
   private get desktopLeftColumnTemplate(): TemplateResult {
     return html`
       <div id="left-column" class="column">
-        ${this.facetTopViewTemplate}
+        <div id="facet-top-view">
+          <slot name="facet-top-slot"></slot>
+        </div>
 
         <div id="facets-header-container">
           <h2 id="facets-header" class="sr-only">Filters</h2>
@@ -2398,10 +2392,6 @@ export class CollectionBrowser
           display: flex;
           margin-top: 10px;
           margin-bottom: 10px;
-          height: var(--facetTopViewHeight, 0px);
-          width: var(--facetTopViewWidth, 0px);
-          color: #fff;
-          border: 1px solid #000;
         }
 
         /**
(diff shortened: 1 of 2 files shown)


src/collection-browser.ts
```diff
@@ -165,6 +165,11 @@ export class CollectionBrowser
    */
   @property({ type: Boolean }) isManageView = false;
 
+  /**
+   * If facet-top view UI active
+   */
+  @property({ type: Boolean }) isFacetTopView = false;
```
nsharma123, line 171: we need to rename these to say isProfiePage or something like this

src/collection-browser.ts
```diff
@@ -493,12 +498,22 @@ export class CollectionBrowser
     `;
   }
 
+  private get facetTopTemplate(): TemplateResult {
+    return html`
+      <div id="facet-top-view" slot="facet-top">
```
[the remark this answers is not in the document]
iisa: yes we want to decouple feature name from behavior at this level.

src/collection-browser.ts
iisa: q: why do we need this boolean at all? the named facet should exist as slot element.

src/collection-browser.ts
```diff
@@ -493,12 +498,22 @@ export class CollectionBrowser
...
+    `;
+  }
+
   /**
    * Template for the desktop version of the left column, displaying the facets sidebar.
    */
   private get desktopLeftColumnTemplate(): TemplateResult {
     return html`
       <div id="left-column" class="column">
+        ${this.isFacetTopView ? this.facetTopTemplate : nothing}
```
iisa, line 515: we talked about just adding the slot here `<slot name='foo></slot>`
whereas in your demo, you toggle the component going into the slot 👍

src/collection-browser.ts
dualcnhq: same as the comment below... were going to use something generic for now to decouple feature name from the actual behavior for now

src/collection-browser.ts
```diff
@@ -2385,6 +2400,19 @@ export class CollectionBrowser
           --leftColumnPaddingRight: 2.5rem;
         }
 
+        #facet-top-view {
+          --facetTopViewHeight: 200px;
+          --facetTopViewWidth: 180px;
```
iisa, line 2405: suggestion: we need to soften this css to remove the JS property.
ideally, this slot will just grow in height when it needs to. perhaps use flex on this? the only requirement is that the slot must show the child's full height. thanks!

also, I see that all of this styling is actually for the demo div, please do remove.

iisa (changes requested):
hi clau, seems we can make this even more simpler by remove that property and ensuring the slot's css is flexible enough to grow given a child's dimensions.

in app-root, when checkbox toggles, it should be inserting that colored div as a direct child of `<collection-browser>` - between lines 371-372 in pic

image

lmk if this makes sense.

iisa (approved):

PR #321, opened 2023-10-26
WEBDEV-6469 Ensure placeholder tiles are not shown if not fetching results

latonv:


src/collection-browser.ts
```diff
@@ -1431,12 +1431,18 @@ export class CollectionBrowser
...
     this.endOfDataReached = false;
-    this.pagesToRender =
-      this.initialPageNumber === 1
-        ? 2 // First two pages are batched into one request when starting from page 1
-        : this.initialPageNumber;
     this.queryErrorMessage = undefined;
 
+    if (!this.baseQuery) {
+      // If we have no query, then there are no pages to fetch anyway
+      this.pagesToRender = 0;
```
iisa, line 1438: q: what is the effect on the draw when this turns to 0?

iisa (approved):
looks good, just 1 question

src/collection-browser.ts
latonv: The intended effect here is that the infinite scroller will not draw any grid/list cells. However, revisiting this it seems to me that there could be an unintended side effect w.r.t. how the current page is determined (possibly leading to `page=Infinity` nonsense). Looking into that a bit further...

PR #322, opened 2023-11-03
WEBDEV-6282 Slot for result CTA tiles in collection-browser

nsharma123:
https://webarchive.jira.com/browse/WEBDEV-6282

Work included:
- User profile page have different tabs where we are showing user specific contents. We have two tabs (uploads and web-archive) from user can directly navigate to upload/create new content.
- Earlier on legacy page, we were showing the Upload/Create tile as first tile in collection-browser for these two tabs. Now we changed mechanism in offshoot where we will show these tiles at the end of the collection-browser.


latonv (changes requested):
Thanks for your work toward getting these CTA tiles in place!

I think we might need another iteration on this to find the right abstraction for how collection browser adds these in.

If the new tiles are fundamentally tied to the parent component's tab structure, then it would likely be more maintainable in the long run to keep that tab-related logic within the parent, instead of bleeding it down into collection browser (which so far is agnostic about what tab it's in).

One approach that might work here is defining a `<slot>` that gets handed to the `<infinite-scroller>` as its last tile of content instead of a `<tile-dispatcher>`. Then we could have a light-DOM structure like this, with the parent defining the content of that slot when needed:

```
<collection-browser>
    <cta-button slot="last-tile"></cta-button>
</collection-browser>
```

and

```
<infinite-scroller>
    <slot name="last-tile" slot="last-tile"></slot>
</infinite-scroller>
```

Note the `<infinite-scroller>` component itself shouldn't require any changes for this to work.

latonv:
As an alternative, if using `<slot>`s feels too cumbersome, we could also consider eliminating the new profile/tab properties in favor of something a bit more generalized and CB-centric, e.g., a property like `ctaTileType: 'none' | 'save' | 'upload' = 'none';`.

That would at least ensure there's a layer of abstraction between the parent component tabs and the CTA tiles, though it would still add to our CB state burden.

PR #325, opened 2023-11-22
WEBDEV-6488 Loans tab navbar custom slot

dualcnhq:


src/collection-browser.ts
```diff
@@ -599,6 +599,15 @@ export class CollectionBrowser
     });
   }
 
+  private get navBarTemplate() {
+    return this.isLoansTab
+      ? html`<slot
+          name="cb-loans-topbar-slot"
+          slot="cb-loans-topbar-slot"
```
nsharma123, line 606: we don't need `slot=""` attribute here.

src/app-root.ts
```diff
@@ -372,6 +372,16 @@ export class AppRoot extends LitElement {
               />
               <label for="enable-facet-top-view">Show facet top view</label>
             </div>
+            <div class="checkbox-control">
+              <input
+                type="checkbox"
+                id="enable-loanstab-topbar-view"
+                @click=${this.loansTabTopBarViewCheckboxChanged}
+              />
+              <label for="enable-facet-top-view">
```
nsharma123, line 381: using different name for, it should be `enable-loanstab-topbar-view`

src/collection-browser.ts
```diff
@@ -617,7 +619,12 @@ export class CollectionBrowser
         @displayModeChanged=${this.displayModeChanged}
         @titleLetterChanged=${this.titleLetterSelected}
         @creatorLetterChanged=${this.creatorLetterSelected}
+        .showLoansTopBar=${this.isLoansTab}
       >
+        <slot
+          name="loans-tab-top-bar-options-slot"
+          slot="loans-tab-top-bar-options-slot"
+        ></slot>
```
nsharma123, line 627: Instead of saying `top-bar` everywhere, we should `filter-bar` to make consistent between offshoot and collection-browser.

nsharma123:
@dualcnhq there are some suggestions for name convention to be consistent with offshoot wording. also please rebase your bbranch with main. there are some conflicts here.

package.json
```diff
@@ -3,7 +3,7 @@
   "description": "The Internet Archive Collection Browser.",
   "license": "AGPL-3.0-only",
   "author": "Internet Archive",
-  "version": "1.14.11",
+  "version": "1.14.11-alpha.10",
```
iisa, line 6: we normally do not commit these alpha tags. as a rule of thumb, alpha version bumps don't really get merged to main...

iisa (approved):
looks good, just had a mention around alpha version hygiene.

package.json
dualcnhq: yup... will remove it

PR #329, opened 2023-12-08
Fixed resizeObserver issue when you are on ?tab=loans

nsharma123:
Currently we are using slot for loans tab sort-bar from offshoot. There are resizeObserver was using on universal sort filter-bar elements which is not present when loans tab. So applied check to make sure the element are there before .addObserver.

This fix is part of https://webarchive.jira.com/browse/WEBDEV-6481 ticket. This issue should be fixed when we added slot for loans tab sort filter-bar.


latonv (approved):
LGTM, though we should separately verify that the loan bar correctly handles resizes too, if it isn't covered by this existing observer.

PR #330, opened 2023-12-11
WEBDEV-6081 Refactoring to better separate data source concerns from UI

latonv:
Factors out the parts of collection browser that handle fetching and storing tile data into a separate data source module, to better separate UI concerns from search behavior.

This is in preparation for collection browser handling search results on multiple tabs of the profile page. A single instance of the component can be used for most or all of these tabs, but only if the fetched data & page state can be cleanly held separately from the component itself. With the separated data source module, the parent page component can manage the data separately and install a chosen data source into its shared collection browser instance as needed when switching between different tabs.

Included are some minor tile adjustments to correctly handle specialized results such as web archive captures.


package.json
```diff
@@ -32,7 +32,7 @@
     "@internetarchive/infinite-scroller": "1.0.1",
     "@internetarchive/local-cache": "^0.2.1",
     "@internetarchive/modal-manager": "^0.2.8",
-    "@internetarchive/search-service": "^1.2.4",
+    "@internetarchive/search-service": "^1.2.5-alpha.6",
```
iisa, line 35: uses PR https://github.com/internetarchive/iaux-search-service/pull/48

src/data-source/collection-browser-data-source.ts
```diff
@@ -0,0 +1,1189 @@
...
+import type {
+  CollectionBrowserSearchInterface,
+  CollectionTitles,
+  PageSpecifierParams,
+} from './models';
+import { sha1 } from '../utils/sha1';
+
+type RequestKind = 'full' | 'hits' | 'aggregations';
+
+export interface CollectionBrowserDataSourceInterface
```
iisa, line 32: suggestion: move the interface to another tile to reduce this file's line length

src/data-source/collection-browser-data-source.ts
```diff
@@ -0,0 +1,1189 @@
...
+
+  /**
+   * Removes all tile models that are currently checked & adjusts the paging
+   * of the data source to account for any new gaps in the data.
+   */
+  removeCheckedTiles(): void;
+}
+
+export class CollectionBrowserDataSource
+  implements CollectionBrowserDataSourceInterface
```
iisa, line 225: Q: what is the overall responsibility for this class - is it to normalize the results data to create view model? can we add 1-2 lines on its usage?

iisa: also i didn't see a discreet unit test for this class - do the current tests catch these updates?

src/data-source/collection-browser-data-source.ts
```diff
@@ -0,0 +1,1189 @@
...
+   */
+  setPageSize(pageSize: number): void {
+    this.reset();
+    this.pageSize = pageSize;
+  }
+
+  /**
+   * @inheritdoc
+   */
+  reset(): void {
```
iisa, line 350: suggestion: raise this higher up the class so we know what a reset data source looks like

src/data-source/collection-browser-data-source.ts
```diff
@@ -0,0 +1,1189 @@
...
+  ): void {
+    this.pages = Object.fromEntries(
+      Object.entries(this.pages).map(([page, tileModels]) => [
+        page,
+        tileModels.map((model, index, array) =>
+          model ? callback(model, index, array) : model
+        ),
+      ])
+    );
+    this.host.requestUpdate();
```
iisa, line 388: Q: is the class that it extends from a lit component?

src/data-source/collection-browser-data-source.ts
```diff
@@ -0,0 +1,1189 @@
...
+    this.map(model => ({ ...model, checked: false }));
+  }
+
+  /**
+   * @inheritdoc
+   */
+  removeCheckedTiles(): void {
+    // To make sure our data source remains page-aligned, we will offset our data source by
+    // the number of removed tiles, so that we can just add the offset when the infinite
+    // scroller queries for cell contents.
```
iisa, line 411: ok

src/data-source/collection-browser-data-source.ts
```diff
@@ -0,0 +1,1189 @@
...
+            if (contentWarning) break;
+          }
+          if (collection === 'no-preview') {
+            contentWarning = true;
+            if (loginRequired) break;
+          }
+        }
+      }
+
+      tiles.push({
```
iisa, line 1050: q: can we do `tiles.push(new TileInfo(result))` and extract this data transform to another file? it'll be easier to test independently

src/data-source/collection-browser-data-source.ts
```diff
@@ -0,0 +1,1189 @@
...
+   * Returns the input string, but removing one set of quotes from all instances of
+   * ""clauses wrapped in two sets of quotes"". This assumes the quotes are already
+   * URL-encoded.
+   *
+   * This should be a temporary measure to address the fact that the __href__ field
+   * sometimes acquires extra quotation marks during query rewriting. Once there is a
+   * full Lucene parser in place that handles quoted queries correctly, this can likely
+   * be removed.
+   */
+  private collapseRepeatedQuotes(str?: string): string | undefined {
```
iisa, line 1123: suggestion: non instance bound functions, let's move to helpers, util area...

src/data-source/collection-browser-data-source.ts
```diff
@@ -0,0 +1,1189 @@
...
+      aggregations: { simpleParams: [filterAggregationKey] },
+      // Fetch all 26 letter buckets
+      aggregationsSize: 26,
+    };
+    params.uid = await this.requestUID(
+      { ...params, sort: sortParams },
+      'aggregations'
+    );
+
+    const searchResponse = await this.host.searchService?.search(
```
iisa, line 1154: Q: the optional chain here denotes that this may not always have a search service - can we explicitly declare that this class needs a searchService to work? it seems like we will always need the service..?

package.json
```diff
@@ -32,7 +32,7 @@
     "@internetarchive/infinite-scroller": "1.0.1",
     "@internetarchive/local-cache": "^0.2.1",
```
iisa, line 33: q: i noticed this was deleted in certain files, can we remove here?

package.json
```diff
@@ -23,7 +24,6 @@
   "types": "dist/index.d.ts",
   "dependencies": {
     "@internetarchive/analytics-manager": "^0.1.2",
-    "@internetarchive/collection-name-cache": "^0.2.16",
```
iisa, line 26: Q: do you think we can deprecate this at the package level?

iisa (commented):
pass completed, lots of good work!
i added some suggestions, questions..

src/data-source/collection-browser-data-source.ts
latonv: I'll add some additional comments around the responsibility & usage, thanks! But yeah, it abstracts out a cleaner model for the server-provided data and the logic determining when that data needs to be modified. There's likely still room for a future pass that improves on the abstraction even further, but I'm wary of any more scope bloat on this ticket.

Many of the existing collection browser unit tests do naturally test all of these data source methods (since CB's own functionality is pretty limited without the data source attached). That said, I can still go in and organize those better. There are probably a few gaps that need filling.

src/data-source/collection-browser-data-source.ts
latonv: The `host` needs to conform to the `ReactiveControllerHost` interface, which requires this method to exist (though it doesn't necessarily have to be a lit component).

src/data-source/collection-browser-data-source.ts
latonv: This piece is migrated over largely unchanged from collection browser, and is one area where we could probably perform a more targeted refactor around page structure. Our approach could use a bit of TLC around sensibly handling pages of data that are smaller than the expected page size (whether due to tiles having been removed, or results omitted on the PPS side).

src/data-source/collection-browser-data-source.ts
latonv: Nice catch, I'll move these out of the data source class!

src/data-source/collection-browser-data-source.ts
latonv: Unfortunately our dependency injection model precludes assuming that app services like this one are always present -- in practice, there is a brief limbo period between when the component is created and when the services are injected, which we account for by triggering an update when the service is passed in.

However, this optional chain in particular is likely unnecessary since we can't reach this point in the method if the search service doesn't exist yet. I'll look at cleaning up the logic here a bit.

package.json
latonv: Possibly! I'll double-check that there isn't anything else using this package...

package.json
latonv: Absolutely! It still needs to be excised from Offshoot itself, but I believe that will be fairly simple -- it really only exists there to be passed into CB anyway.

iisa: ok, here is the ticket for it: https://webarchive.jira.com/browse/WEBDEV-6594

package.json
iisa: Here is ticket for this: https://webarchive.jira.com/browse/WEBDEV-6595

iisa (approved):
hi, things look good on my end. re: addressing suggestions, we can do that in smaller chunks.. the bulk of this work looks good to go on my part 👍 @latonv

iisa (changes requested):
unit tests not working as expected..

src/data-source/collection-browser-data-source.ts
```diff
@@ -0,0 +1,1162 @@
...
+    // pages 1 and 2 together in a single request).
+    const numPages = pageNumber === 1 ? numInitialPages : 1;
+    const numRows = this.pageSize * numPages;
+
+    // if a fetch is already in progress for this query and page, don't fetch again
+    const { pageFetchQueryKey } = this;
+    const pageFetches =
+      this.pageFetchesInProgress[pageFetchQueryKey] ?? new Set();
+    if (pageFetches.has(pageNumber)) {
+      console.log(
```
nsharma123, line 947: we can introduce `log()` inside collection-browser as well for logs on dev instances.

latonv: Thanks, I've ported in the `log()` util function from Offshoot. Most of the log statements are removed now, but a few that may be helpful for further debugging now use this function instead of bare `console.log`.

PR #331, opened 2023-12-12
Add prop to suppress facets pane & associated requests

latonv:
Adds a `suppressFacets` property to collection-browser to allow quick disabling of the facets pane and its associated PPS requests.


iisa (approved):
:lgtm:

PR #336, opened 2024-01-03
Unit Tests Updates

iisa:
- Fix Test `<sort-filter-bar>`: Resize Observer breaks when toggling between default view & custom slotted view
- Mute Tests `<collection-browser>`:
  - `removeCheckedItems` are not working as expected from unit test level
  - `placeholderType` not setting properly
  - skipped a couple of tests because elemental order has broken some

### DIFF
--- a/src/data-source/collection-browser-data-source.ts
+++ b/src/data-source/collection-browser-data-source.ts
@@ -13,6 +13,7 @@ import {
   SearchType,
 } from '@internetarchive/search-service';
 import type { MediaType } from '@internetarchive/field-parsers';
+import type { PageElementName } from '@internetarchive/search-service/dist/src/responses/page-elements';
 import {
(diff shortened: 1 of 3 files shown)


src/sort-filter-bar/sort-filter-bar.ts
```diff
@@ -216,28 +216,37 @@ export class SortFilterBar
   private disconnectResizeObserver(
     resizeObserver: SharedResizeObserverInterface
   ) {
-    resizeObserver.removeObserver({
-      target: this.sortSelectorContainer,
-      handler: this,
-    });
+    if (this.sortSelectorContainer) {
```
nsharma123, line 219: I have covered this part in one of my earlier PR https://github.com/internetarchive/iaux-collection-browser/pull/329. we needs to rebase @latonv original branch for this fix.

PR #337, opened 2024-01-04
Demo: add placeholder views

iisa:


src/collection-browser.ts
```diff
@@ -263,9 +263,16 @@ export class CollectionBrowser
      *
      * We disable it during the automated scroll since we don't want to fetch pages for intervening cells the
      * user may never see.
+     *
+     * No need to fetch if placeholder is being displayed
      */
-    if (!model && !this.isScrollingToCell) {
+    if (!this.placeholderType && !model && !this.isScrollingToCell) {
```
iisa, line 269: this block fires even though we render the placeholder. this reduces fetch checks since we request details for each result...

src/app-root.ts
```diff
@@ -395,6 +395,55 @@ export class AppRoot extends LitElement {
               </label>
             </div>
           </fieldset>
+          <fieldset class="Placeholder-type">
+            <legend>Placeholder type</legend>
+            <div class="checkbox-control">
+              <input
```
nsharma123, line 401: I would make them radiobutton group which match with other checkbox styling :)

src/app-root.ts
```diff
@@ -489,6 +538,50 @@ export class AppRoot extends LitElement {
     }
   }
 
+  private async showEmptyPlaceholder(placeholderType: string) {
+    switch (placeholderType) {
+      case 'empty query':
```
nsharma123, line 543: I know we are on demo for developers but we should not use string slugs for checking conditions. I would write these something like `empty-query`, `empty-collection`, `empty-profile` something like this.

src/collection-browser.ts
```diff
@@ -263,9 +263,16 @@ export class CollectionBrowser
...
       const pageNumber = Math.floor(offsetIndex / this.pageSize) + 1;
+      console.warn('****** FETCH PAGE tileModelAtCellIndex', {
```
nsharma123, line 271: can you please our custom log() function to render logs messages for development purpose?

src/collection-browser.ts
```diff
@@ -995,7 +1002,15 @@ export class CollectionBrowser
     this.restoreState();
   }
 
+  willUpdate() {
+    this.setPlaceholderType();
```
nsharma123, line 1006: Just curious, whe `willUpdate()` function executes? (still learning)

src/collection-browser.ts
```diff
@@ -1379,11 +1394,10 @@ export class CollectionBrowser
     )
       return;
 
-    console.log(
-      'CB will reset',
-      this.baseQuery,
-      JSON.stringify(this.selectedFacets)
-    );
+    console.log('CB will reset', {
```
nsharma123, line 1397: let's add our `log()` util function here.

src/collection-browser.ts
iisa: yes I can remove these, actually once we get stable work on offshoot 👍

src/app-root.ts
iisa: I tried to even do radio buttons but they wouldn't toggle, so I figured this is the easiest way for me to get something on the board that works. 👍 I will hold as buttons for now as I am closer to main behavioural fix, this demo helped me come up with creating dynamic interstitial...

PR #342, opened 2024-01-27
WEBDEV-6279 Review snippets on tiles

latonv:
Adds review snippets showing a preview of a patron's item review on grid & extended list tiles.


bfalling (approved):
LGTM

PR #343, opened 2024-01-29
WEBDEV-6637 Ensure sort dropdown selections are sticky and correctly reflect active defaults

latonv:
This PR tweaks the sort selection dropdowns for Views/Dates so that:
1. After making a dropdown selection, changing the sort to something outside of that dropdown will leave the last selected option visible (i.e., "sticky" selection), instead of resetting the dropdown to its default state.
2. When no dropdown selection has been made yet (so the default option is visible on the dropdown label), opening the dropdown menu will correctly show that default option as the selected menu item. (Without this fix, the menu is opening in an indeterminate state with nothing selected.)

### DIFF
--- a/src/sort-filter-bar/sort-filter-bar.ts
+++ b/src/sort-filter-bar/sort-filter-bar.ts
@@ -167,6 +167,12 @@ export class SortFilterBar
   }
 
   willUpdate(changed: PropertyValues) {
+    console.log(
+      'sort bar updated:\n',
+      [...changed.entries()]
+        .map(([k, v]) => `${String(k)}: ${v} -> ${this[k as keyof this]}`)
+        .join('\n')
+    );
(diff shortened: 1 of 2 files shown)


src/sort-filter-bar/sort-filter-bar.ts
```diff
@@ -156,16 +166,41 @@ export class SortFilterBar
     `;
   }
 
+  willUpdate(changed: PropertyValues) {
+    console.log(
```
mc2, line 170: Question: Are we planning to leave this `console.log` in?
Suggestion: `// eslint-disable-next-line no-console`

latonv: Oops, that can be removed. Thanks!

mc2 (approved):

PR #348, opened 2024-02-03
WEBDEV-6640 Generalize loan bar slot to permit broader customization

latonv:
Replaces the existing `isLoansTab` property with a more general `enableSortOptionsSlot` property that performs the same role, albeit with a purpose that is transparently more general. This option, like `isLoansTab` before it, serves only to indicate that a customizable slot should be rendered in the sort bar in place of the usual sorting options.

The sort bar CSS is slightly tweaked to ensure that the layout does not have undesired changes when this option is enabled without assigning anything to the resulting slot -- which is the case when we only wish to omit the sort options without replacement, as on the Web Archives tab of the profile page.

### DIFF
--- a/src/collection-browser.ts
+++ b/src/collection-browser.ts
@@ -171,7 +171,8 @@ export class CollectionBrowser
    */
   @property({ type: Boolean }) isManageView = false;
 
-  @property({ type: Boolean }) isLoansTab = false;
+  /** Whether to replace the default sort options with a slot for customization (default: false) */
+  @property({ type: Boolean }) enableSortOptionsSlot = false;
 
   /**
    * The results per page so we can paginate.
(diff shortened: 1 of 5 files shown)


src/collection-browser.ts
```diff
@@ -171,7 +171,8 @@ export class CollectionBrowser
    */
   @property({ type: Boolean }) isManageView = false;
 
-  @property({ type: Boolean }) isLoansTab = false;
+  /** Whether to replace the default sort options with a slot for customization (default: false) */
+  @property({ type: Boolean }) enableSortOptionsSlot = false;
```
iisa, line 175: can we have it reflect as well onto the element as attr

iisa (approved):
thanks! looks good.

PR #349, opened 2024-02-06
open analytics handler

iisa:
when adding property with JS and not through lit template, we need property to be public.

[see job](https://git.archive.org/www/offshoot/-/jobs/719176)
image


latonv (approved):
Thanks for catching & fixing this!

PR #350, opened 2024-02-10
WEBDEV-6292 Adjust sort bar slots/props to accommodate inline search bar

latonv:
Adds a slot for the right side of the sort options, and adds a property to suppress rendering the display mode options. Both of these are to allow a search bar to be slotted in which can be expanded to the full width of the bar on mobile as needed.

### DIFF
--- a/src/app-root.ts
+++ b/src/app-root.ts
@@ -376,10 +376,20 @@ export class AppRoot extends LitElement {
               <input
                 type="checkbox"
                 id="enable-sortbar-left-slot"
-                @click=${this.sortBarSlotCheckboxChanged}
+                @click=${this.sortBarLeftSlotCheckboxChanged}
               />
               <label for="enable-sortbar-left-slot"
                 >Show sortbar left slot</label
+              >
+            </div>
+            <div class="checkbox-control">
+              <input
+                type="checkbox"
+                id="enable-sortbar-right-slot"
+                @click=${this.sortBarRightSlotCheckboxChanged}
+              />
+              <label for="enable-sortbar-right-slot"
+                >Show sortbar right slot</label
               >
             </div>
             <div class="checkbox-control">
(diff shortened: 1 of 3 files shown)


src/app-root.ts
```diff
@@ -733,25 +743,50 @@ export class AppRoot extends LitElement {
...
+      div.style.setProperty('width', '15rem');
+      div.style.backgroundColor = '#000';
+      div.setAttribute('slot', 'sort-options-right');
+
       this.collectionBrowser.appendChild(div);
     } else {
-      this.collectionBrowser.removeChild(
-        this.collectionBrowser.lastElementChild as Element
+      const slottedEl = this.collectionBrowser.querySelector(
+        '[slot="sort-options-right"]'
```
latonv, line 787: Cleaned up some dubious logic in these demo app methods. Since multiple elements might be slotted in at once, we can't rely on the last element being the one that should be removed.

src/collection-browser.ts
```diff
@@ -136,6 +136,8 @@ export class CollectionBrowser
 
   @property({ type: Boolean }) suppressSortBar = false;
 
+  @property({ type: Boolean }) suppressDisplayModes = false;
```
iisa, line 139: idk how I feel about a new property, can we just do it in the resize callback and save us a new property at the top level?

iisa (approved):
thanks looks good, added some questions around new property & how to see all the demo changes + how to test fix on demo

src/collection-browser.ts
latonv: Unfortunately the use case here isn't tied to a specific sort bar width like the sort options. We want these display modes to disappear when a patron has expanded the search bar on the profile page, and re-appear when they focus elsewhere, regardless of the width of the page or sort bar container.

I'm no fan of adding new properties, but I think in this case we do need one for the sake of passing UI state. Perhaps a later refactoring pass should attempt to consolidate some of our various boolean properties into a single options object.

PR #351, opened 2024-02-13
Customized manage views label while removing items

nsharma123:
Currently when you remove items from Lists tab, there is default label in manage-bar says `Select items to un-favorite`.

Let's just make it customization while using `<collection-browser>` so that we can pass custom message as well since we are going to use this multiple locations like removing fav- item from fav- collection, remove lists, uploads, collections etc.. from new profile page.

Reference to https://webarchive.jira.com/browse/WEBDEV-6680

### DIFF
--- a/src/manage/manage-bar.ts
+++ b/src/manage/manage-bar.ts
@@ -14,7 +14,7 @@ export class ManageBar extends LitElement {
   /**
    * The label displayed in front of the management buttons
    */
-  @property({ type: String }) label = msg('Select items to un-favorite');
+  @property({ type: String }) label = msg('Select items to remove');
 
   /**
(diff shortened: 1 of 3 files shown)


src/manage/manage-bar.ts
```diff
@@ -84,7 +84,7 @@ export class ManageBar extends LitElement {
         display: flex;
         align-items: center;
         column-gap: 5px;
-        padding: 10px 0 20px;
+        padding: 20px 0 20px;
```
nsharma123, line 87: Prevent to jumping `collection-browser/tiles` when toggle manageView.

latonv (approved):
LGTM

PR #353, opened 2024-02-16
Collection-Browser - toggle the Remove button in ManageView while while removing items

nsharma123:
https://webarchive.jira.com/browse/WEBDEV-6702

when you open manageView, the delete button should be disable initially. when you select items, it should turn to active state.

### DIFF
--- a/src/collection-browser.ts
+++ b/src/collection-browser.ts
@@ -646,30 +646,19 @@ export class CollectionBrowser
   private get manageBarTemplate(): TemplateResult {
     return html`
       <manage-bar
-        .label=${this.manageViewLabel}
-        ?showSelectAll=${true}
-        ?showUnselectAll=${true}
-        ?enableRemoveButton=${this.manageViewRemoveButtonEnable}
+        label=${this.manageViewLabel}
+        showSelectAll
+        showUnselectAll
+        ?enableRemoveButton=${this.dataSource.checkedTileModels.length !== 0}
         @removeItems=${this.handleRemoveItems}
-        @selectAll=${() => {
-          this.dataSource.checkAllTiles();
-          this.toggleManageViewRemoveButtonState();
-        }}
-        @unselectAll=${() => {
-          this.dataSource.uncheckAllTiles();
-          this.toggleManageViewRemoveButtonState();
-        }}
+        @selectAll=${() => this.dataSource.checkAllTiles()}
+        @unselectAll=${() => this.dataSource.uncheckAllTiles()}
         @cancel=${() => {
           this.isManageView = false;
           this.dataSource.uncheckAllTiles();
         }}
       ></manage-bar>
     `;
-  }
-
-  private toggleManageViewRemoveButtonState() {
-    this.manageViewRemoveButtonEnable =
-      !!this.dataSource.checkedTileModels.length;
   }
 
   /**
@@ -1695,7 +1684,7 @@ export class CollectionBrowser
       const cellIndex = this.dataSource.indexOf(event.detail);
       if (cellIndex >= 0) this.infiniteScroller?.refreshCell(cellIndex);
 
-      this.toggleManageViewRemoveButtonState();
+      this.requestUpdate();
     }
 
     this.analyticsHandler?.sendEvent({
--- a/src/collection-browser.ts
+++ b/src/collection-browser.ts
@@ -221,8 +221,6 @@ export class CollectionBrowser
 
   @state() private placeholderType: PlaceholderType = null;
 
-  @state() private manageViewRemoveButtonEnable = false;
-
   @query('#content-container') private contentContainer!: HTMLDivElement;
 
   @query('#left-column') private leftColumn?: HTMLDivElement;
--- a/src/collection-browser.ts
+++ b/src/collection-browser.ts
@@ -221,6 +221,8 @@ export class CollectionBrowser
 
   @state() private placeholderType: PlaceholderType = null;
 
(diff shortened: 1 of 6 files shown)


src/collection-browser.ts
```diff
@@ -658,6 +667,11 @@ export class CollectionBrowser
     `;
   }
 
+  private toggleManageViewRemoveButtonState() {
```
mc2, line 670: Suggestion: method name: This doesn't "toggle" the button state, it more precisely "updates" it.

Suggest replace state variable with method
```
\\ ?enableRemoveButton=${this.manageViewRemoveButtonEnable}
?enableRemoveButton=${this.anySelected}
...
get anySelected(): boolean {
  return this.dataSource.checkedTileModels.length !== 0;
}
```

src/collection-browser.ts
```diff
@@ -658,6 +667,11 @@ export class CollectionBrowser
...
+    this.manageViewRemoveButtonEnable =
+      !!this.dataSource.checkedTileModels.length;
```
mc2, line 672: Suggestion: When I use `!!` I get a linting error. If I'm correct, TS style suggests using the same type for comparisons

src/collection-browser.ts
```diff
@@ -644,12 +646,19 @@ export class CollectionBrowser
...
-        showSelectAll
-        showUnselectAll
+        .label=${this.manageViewLabel}
+        ?showSelectAll=${true}
+        ?showUnselectAll=${true}
+        ?enableRemoveButton=${this.manageViewRemoveButtonEnable}
         @removeItems=${this.handleRemoveItems}
-        @selectAll=${() => this.dataSource.checkAllTiles()}
-        @unselectAll=${() => this.dataSource.uncheckAllTiles()}
+        @selectAll=${() => {
```
mc2, line 654: Per comment below: if you use a method rather than a state for `enableRemoveButton`, this change is unnecessary.

test/manage/manage-bar.test.ts
```diff
@@ -45,6 +45,13 @@ describe('Manage bar', () => {
     expect(el.shadowRoot?.querySelector('.unselect-all-btn')).to.exist;
   });
 
+  it('have delete button disabled', async () => {
+    const el = await fixture<ManageBar>(
+      html`<manage-bar ?enableRemoveButton=${false}></manage-bar>`
```
latonv, line 50: Note `?myattr=${false}` is a no-op in most contexts (it just means "don't include a `myattr` attribute on the element"). But I suppose it's okay here just as an extra visual indication of what's being tested.

latonv (commented):
Thanks for working to improve the manage bar UX! I've noted what I think are some ways to reduce state complexity here, and a tweak to the button disabling that should have accessibility benefits.

src/manage/manage-bar.ts
```diff
@@ -146,6 +154,15 @@ export class ManageBar extends LitElement {
         background: #d2322d;
         border: 1px solid #ac2925;
       }
+
+      .remove-btn.disable {
+        opacity: 0.5;
```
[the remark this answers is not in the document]
nsharma123: We have a custom build ia-button.ts styling for all the button across petabox and other repos. It has defined some sort of button having different states. We are using this styling on multiple places right now. like item-navigator, ia-book-actions, offshoot, iaux-acccount-settings etc...

src/collection-browser.ts
```diff
@@ -647,6 +647,7 @@ export class CollectionBrowser
         label=${this.manageViewLabel}
         showSelectAll
         showUnselectAll
+        ?enableRemoveButton=${this.dataSource.checkedTileModels.length !== 0}
```
iisa, line 650: suggestion: rename to `removeAllowed` so that we can get closer to the responsibility.

`enableRemoveButton` => `removeAllowed`

src/collection-browser.ts
```diff
@@ -1680,6 +1681,8 @@ export class CollectionBrowser
       // this.mapDataSource(model => ({ ...model }));
       const cellIndex = this.dataSource.indexOf(event.detail);
       if (cellIndex >= 0) this.infiniteScroller?.refreshCell(cellIndex);
+
+      this.requestUpdate();
```
iisa, line 1685: ok

src/styles/ia-button.ts
```diff
@@ -0,0 +1,107 @@
+import { css } from 'lit';
```
iisa, line 1: lol let's package this up 🎁

test/manage/manage-bar.test.ts
```diff
@@ -45,6 +45,11 @@ describe('Manage bar', () => {
     expect(el.shadowRoot?.querySelector('.unselect-all-btn')).to.exist;
   });
 
+  it('have delete button disabled', async () => {
+    const el = await fixture<ManageBar>(html`<manage-bar></manage-bar>`);
+    expect(el.shadowRoot?.querySelector('.remove-btn:disabled')).to.exist;
+  });
```
iisa, line 51: let's add another test here in this block when the `removeAllowed` prop is toggled to complete

iisa (approved):
thanks! looks good, had some suggestions. once applied, lgtm

src/styles/ia-button.ts
nsharma123: Yeah, we are using this single file in multiple repo and sometimes make specific changes in this file per requirements leads to inconsistency in this file. we should publish as a package so we can have atleast common behaviour on this file.

PR #355, opened 2024-02-21
WEBDEV-6705 Fire analytics when mobile facets toggled

latonv:
Adds an analytics event that fires whenever the mobile facets accordion is opened or closed.


mc2 (approved):
LGTM

PR #357, opened 2024-02-28
WEBDEV-6697 Extract session context from search service responses

latonv:
Responses from the PPS generally include a `session_context` branch containing info about the user session within which the response was generated. Notably, it includes whether the user has privs which should permit them to perform full-text searches even in the absence of public (i.e., not `noindex`) items with searchable text.

This PR adds support for that session context and makes it available in the data source.

### DIFF
--- a/test/mocks/mock-search-responses.ts
+++ b/test/mocks/mock-search-responses.ts
@@ -31,6 +31,7 @@ export const getMockSuccessSingleResult: () => Result<
       },
     },
     rawResponse: {},
+    sessionContext: {},
     response: {
       totalResults: 1,
       returnedCount: 1,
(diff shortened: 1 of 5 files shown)


test/mocks/mock-search-responses.ts
```diff
@@ -31,6 +31,7 @@ export const getMockSuccessSingleResult: () => Result<
       },
     },
     rawResponse: {},
+    sessionContext: {},
```
latonv, line 34: Just updating our mock responses to include empty session context, for type correctness & completeness' sake.

iisa: I appreciate that bc then things match

iisa (approved):
lgtm!

PR #361, opened 2024-03-06
WEBDEV-6738 Propagate profile params to more-facets component

latonv:
Ensures that profile-related params are propagated all the way through to the `<more-facets-content>` component to be used when fetching full facet lists.

### DIFF
--- a/src/collection-browser.ts
+++ b/src/collection-browser.ts
@@ -932,9 +932,7 @@ export class CollectionBrowser
         @histogramDateRangeUpdated=${this.histogramDateRangeUpdated}
         .collectionPagePath=${this.collectionPagePath}
         .parentCollections=${this.dataSource.parentCollections}
-        .withinCollection=${this.withinCollection}
-        .withinProfile=${this.withinProfile}
-        .profileElement=${this.profileElement}
+        .pageSpecifierParams=${this.dataSource.pageSpecifierParams}
         .searchService=${this.searchService}
         .featureFeedbackService=${this.featureFeedbackService}
         .recaptchaManager=${this.recaptchaManager}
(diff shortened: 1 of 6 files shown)


src/collection-browser.ts
```diff
@@ -932,7 +932,7 @@ export class CollectionBrowser
         @histogramDateRangeUpdated=${this.histogramDateRangeUpdated}
         .collectionPagePath=${this.collectionPagePath}
         .parentCollections=${this.dataSource.parentCollections}
-        .withinCollection=${this.withinCollection}
+        .pageSpecifierParams=${this.dataSource.pageSpecifierParams}
```
iisa, line 935: 👍

src/collection-facets/more-facets-content.ts
```diff
@@ -129,28 +131,30 @@ export class MoreFacetsContent extends LitElement {
     }
   }
 
+  /**
+   * Whether facet requests are for the search_results page type (either defaulted or explicitly).
+   */
+  private get isSearchResultsPage(): boolean {
+    return [undefined, 'search_results'].includes(
```
iisa, line 138: q: why not === ?

s: pageType, can we make that a type something for all to share? and also so we know what's allowed-ish

iisa (approved):
looks good, just had some thinking around pageType

src/collection-facets/more-facets-content.ts
latonv: Ah yeah, `search_results` is the default page type so we need to check against `undefined` here too. There's a `PageType` type we can use from the search service here, so I'll update the typing of this array. Thanks!

bfalling (approved):
LGTM. No obvious blockers.

src/collection-facets/more-facets-content.ts
latonv: Updated for clarity & type correctness

PR #367, opened 2024-03-13
WEBDEV-6753 Allow facets to be lazy-loaded on demand in mobile view

latonv:
Adds a bit of logic to the data source so that facets can be lazy-loaded when they are actually needed, avoiding expensive facet requests in cases where they are not (i.e., the majority of visits).


dualcnhq (approved):
:LGTM:

PR #368, opened 2024-03-25
WEBDEV-6756 Update total result count when checked tiles are removed

latonv:
When removing checked tiles via collection-browser's management UI, currently the total result count next to the facets does not change. This PR fixes that bug to ensure the total result count reflects the removals immediately.


mc2 (approved):
Looks okay. 

Assuming we're not re-querying for updated values, and not clear where the tile model removal from data store happens. Not important for this review.

PR #372, opened 2024-05-01
Prevent page number from incrementing early on very tall screens

latonv:
Currently if a visitor is using a tall/vertical screen that is capable of showing more than a single page of results at once, they will erroneously have a `page` param added to the URL that is one page ahead of their actual position. Most egregiously, this causes a `page=2` param to appear when they are at the top of the page, before scrolling at all.

This PR ensures that the current page indicated in the URL is never more than one page beyond that of the first visible cell.


mc2 (approved):
Works and looks great

PR #376, opened 2024-05-10
WEBDEV-6853 Opt-in facet loading for desktop view

latonv:
Consolidates the `suppressFacets` and `lazyLoadFacets` options into a single property for adjusting the facet loading strategy, adding an additional option to display opt-in style facets in desktop view as well. This new option delays the facet load until patrons open a disclosure widget, much like in the current mobile behavior. The intent is for this new option to be used when it is necessary to reduce facet load on the backend temporarily.

### DIFF
--- a/src/collection-browser.ts
+++ b/src/collection-browser.ts
@@ -43,6 +43,7 @@ import {
   sortOptionFromAPIString,
   SORT_OPTIONS,
   defaultProfileElementSorts,
+  FacetLoadStrategy,
 } from './models';
 import {
   RestorationStateHandlerInterface,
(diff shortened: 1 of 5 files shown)


src/collection-browser.ts
```diff
@@ -43,6 +43,7 @@ import {
   sortOptionFromAPIString,
   SORT_OPTIONS,
   defaultProfileElementSorts,
+  FacetLoadStrategy,
```
nsharma123, line 46: @latonv here we are following different case styling for all the models. Somewhere we using all caps or PascalCase or camelCase. can we please use one standard for these? we can just rename other as well in this PR?

latonv: I don't think it would make sense for these all to use the same casing, because they represent different things that are _meant_ to be cased differently in our conventions. For instance, `FacetLoadStrategy` is a type, while `sortOptionFromAPIString` is a function, and `SORT_OPTIONS` is an exported module-level constant. These tend to be distinguished visually by giving them different casing so that it is clearer at a glance what the identifier represents. Some further reading: https://google.github.io/styleguide/tsguide.html#naming-rules-by-identifier-type

One thing that might help, though, is re-organizing some of these exports into separate, smaller modules so that we don't have all of the above coming from the same place. (That project probably warrants a separate PR)

bfalling (approved):
LGTM. No obvious blockers.

PR #380, opened 2024-06-20
WEBDEV-6909 Ensure radio searches are properly read/written to URL

latonv:
Adds support in the restoration state handler for `sin=RADIO` being read from/written to the URL. This is to support radio searches being performed entirely within Offshoot rather than redirecting away to a legacy page.


iisa (approved):
👍

PR #381, opened 2024-06-26
WEBDEV-6333 Don't move facets around when clicked in More... dialog

latonv:
Currently facets in the More... dialog have the following behavior: when checked they immediately bubble up to the top of the page, and when unchecked they immediately return to their original sorted position. This behavior is similar to how facet selections bubble up to the top when selected in the _sidebar_, although the sidebar triggers a full refresh upon each selection while the dialog does not.

While the original intent was presumably to keep all the selected facets grouped separately from the unselected ones, the reflow upon making any facet changes is jarring and produces frustrating UX in many cases.

This PR adjusts the behavior in the dialog to leave facets in place when they are clicked, and performs a bit of cleanup & refactoring to improve some unusual data flows in our facet components that were making changes more challenging than they needed to be.

Moreover, facets that were selected/hidden prior to opening the dialog now appear only once, at the very beginning of the pagination, rather than following you from page to page (which had some undesired behavior in cases where many facets were selected).

### DIFF
--- a/src/collection-facets/facets-template.ts
+++ b/src/collection-facets/facets-template.ts
@@ -8,75 +8,19 @@ import {
 } from 'lit';
 import { customElement, property } from 'lit/decorators.js';
 import { repeat } from 'lit/directives/repeat.js';
-import {
-  FacetGroup,
-  FacetBucket,
-  SelectedFacets,
-  getDefaultSelectedFacets,
-  FacetEventDetails,
-} from '../models';
+import type { FacetGroup, FacetBucket, FacetEventDetails } from '../models';
 import type { CollectionTitles } from '../data-source/models';
-import { FacetRow } from './facet-row';
+import './facet-row';
 
 @customElement('facets-template')
 export class FacetsTemplate extends LitElement {
   @property({ type: Object }) facetGroup?: FacetGroup;
 
-  @property({ type: Object }) selectedFacets?: SelectedFacets;
-
-  @property({ type: String }) renderOn?: string;
-
   @property({ type: Object })
   collectionTitles?: CollectionTitles;
 
   private facetClicked(e: CustomEvent<FacetEventDetails>) {
-    const { bucket, negative } = e.detail;
-    if (bucket.state === 'none') {
-      this.facetUnchecked(bucket);
-    } else {
-      this.facetChecked(bucket, negative);
-    }
-
     this.dispatchFacetClickEvent(e.detail);
-  }
-
-  private facetChecked(bucket: FacetBucket, negative: boolean) {
-    const { facetGroup, selectedFacets } = this;
-    if (!facetGroup) return;
-
-    let newFacets: SelectedFacets;
-    if (selectedFacets) {
-      newFacets = {
-        ...selectedFacets,
-      };
-    } else {
-      newFacets = getDefaultSelectedFacets();
-    }
-    newFacets[facetGroup.key][bucket.key] = {
-      ...bucket,
-      state: FacetRow.getFacetState(true, negative),
-    };
-
-    this.selectedFacets = newFacets;
-    this.dispatchSelectedFacetsChanged();
-  }
-
-  private facetUnchecked(bucket: FacetBucket) {
-    const { facetGroup, selectedFacets } = this;
-    if (!facetGroup) return;
-
-    let newFacets: SelectedFacets;
-    if (selectedFacets) {
-      newFacets = {
-        ...selectedFacets,
-      };
-    } else {
-      newFacets = getDefaultSelectedFacets();
-    }
-    delete newFacets[facetGroup.key][bucket.key];
-
-    this.selectedFacets = newFacets;
-    this.dispatchSelectedFacetsChanged();
   }
 
   private dispatchFacetClickEvent(detail: FacetEventDetails) {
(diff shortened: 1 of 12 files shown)


src/collection-facets/facets-template.ts
```diff
@@ -8,77 +8,21 @@ import {
...
-import { FacetRow } from './facet-row';
+import './facet-row';
 
 @customElement('facets-template')
 export class FacetsTemplate extends LitElement {
   @property({ type: Object }) facetGroup?: FacetGroup;
 
-  @property({ type: Object }) selectedFacets?: SelectedFacets;
-
-  @property({ type: String }) renderOn?: string;
```
latonv, line 27: I've sort of gutted this component in order to eliminate some anti-patterns:

1. Inverted update flow (`selectedFacetsChanged` etc). This component's prior behavior was to accept the `selectedFacets` state, modify that state non-immutably, and then emit an event notifying parents of the state change. This felt somewhat backwards. The best practice we aim to follow is pushing state changes downwards, not upwards -- low-level components should emit events that allow their ancestors to update the state appropriately and flow any changes back down. So, I've migrated the parents to listen for the `facetClick` event instead, eliminating any need for `selectedFacets` here because this component has no business modifying that high-level state directly.

3. Explicitly caring where it's rendered (`renderOn`). This component was handling all the logic for how to change its styles according to where it was being used (`page` vs `modal`). This is better handled by just exposing some CSS variables and letting each parent define how to customize them -- this component shouldn't have to care about those details.

src/collection-facets/more-facets-content.ts
```diff
@@ -45,59 +48,91 @@ import {
...
+  mergeSelectedFacets,
+  sortBucketsBySelectionState,
+  updateSelectedFacetBucket,
+} from '../utils/facet-utils';
 
 @customElement('more-facets-content')
 export class MoreFacetsContent extends LitElement {
   @property({ type: String }) facetKey?: FacetOption;
 
-  @property({ type: String }) facetAggregationKey?: FacetOption;
```
latonv, line 53: This property was redundant.

src/collection-facets/more-facets-content.ts
```diff
@@ -45,59 +48,91 @@ import {
...
-  @property({ type: Object, attribute: false })
-  analyticsHandler?: AnalyticsManagerInterface;
-
-  @state() aggregations?: Record<string, Aggregation>;
+  @property({ type: Object }) modalManager?: ModalManagerInterface;
 
-  @state() facetGroup?: FacetGroup[] = [];
+  @property({ type: Object }) searchService?: SearchServiceInterface;
 
-  @state() facetGroupTitle?: String = '';
```
latonv, line 82: This property was also redundant -- the `facetGroup` already contains its title as a property.

src/collection-facets/more-facets-content.ts
```diff
@@ -45,59 +48,91 @@ import {
...
+  @state() private facetGroup?: FacetGroup;
 
-  @state() paginationSize = 0;
+  /**
+   * An object holding any changes the patron has made to their facet selections
+   * within the modal dialog but which they have not yet applied. These are
+   * eventually merged into the existing `selectedFacets` when the patron applies
+   * their changes, or discarded if they cancel/close the dialog.
+   */
+  @state() private unappliedFacetChanges = getDefaultSelectedFacets();
```
latonv, line 114: New facet selections within the modal are saved here across different pages of facets, and only merge them in later when the patron chooses to apply them.

src/collection-facets/more-facets-content.ts
```diff
@@ -195,32 +229,24 @@ export class MoreFacetsContent extends LitElement {
   /**
    * Combines the selected facets with the aggregations to create a single list of facets
    */
-  private get mergedFacets(): FacetGroup[] | void {
```
latonv, line 198: It seems like a great deal of this logic was originally copied over from `collection-facets.ts`, but wasn't fully adjusted to account for the differences (e.g., More Facets only ever showing a single facet group, instead of multiple at once).

I've aimed to simplify the logic around these getters to (1) eliminate a bunch of unnecessary looping over aggregation types that aren't relevant to the one being displayed, and (2) return only the relevant facet group, rather than an array of groups where only one is useful.

src/collection-facets/more-facets-content.ts
```diff
@@ -470,18 +525,10 @@ export class MoreFacetsContent extends LitElement {
...
       css`
-        @media (max-width: 560px) {
-          section#more-facets {
-            max-height: 450px;
-          }
-          .facets-content {
-            overflow-y: auto;
-            height: 300px;
-          }
-        }
```
latonv, line 481: Moved these media queries down below, because otherwise the unconditional rules below them took precedence.

src/utils/facet-utils.ts
```diff
@@ -0,0 +1,152 @@
+import {
+  FacetOption,
+  getDefaultSelectedFacets,
+  type FacetBucket,
+  type SelectedFacets,
+} from '../models';
+
+// This file contains several helper functions designed to make working immutably
+// with SelectedFacet objects and FacetGroups cleaner & easier.
```
latonv, line 9: I split off several helper functions to deal with some common facet-related actions we perform across various components.

test/collection-facets/facets-template.test.ts
```diff
@@ -66,14 +52,12 @@ describe('Render facets', () => {
...
 
-  it('emits facet click and change events when a facet is selected/deselected', async () => {
+  it('emits facet click event when a facet is selected/deselected', async () => {
     const facetClickSpy = sinon.spy();
-    const facetChangeSpy = sinon.spy();
     const el = await fixture<FacetsTemplate>(
       html`<facets-template
         .facetGroup=${facetGroup}
         @facetClick=${facetClickSpy}
-        @selectedFacetsChanged=${facetChangeSpy}
```
latonv, line 76: This event no longer exists on the `facet-template` component -- tests updated accordingly.

iisa (approved):
thanks for the coverage while making good sweeping changes!

PR #382, opened 2024-07-05
WEBDEV-6936 Adding test-id for playwright tests

dualcnhq:


latonv (approved):
I think I would like to learn more about the problem these testids are trying to solve, but from my reading of the Playwright docs it seems like this is one of the recommended approaches to avoid brittle test dependencies on specific DOM structure.

One thing that would be helpful is documenting in comments why each marked element needs a `testid` (i.e., why is it not possible to find this element by role?). I would like it to be easy for us to maintain this code with a clear understanding of how to adjust these if necessary. If we can add that, LGTM!

PR #383, opened 2024-07-09
WEBDEV-6944 Add `opt-in-or-login` facet load strategy and related bugfix

latonv:
This PR adds a new `opt-in-or-login` facet load strategy which works like `opt-in` for anyone who is not logged into an account, but like `eager` for anyone who is logged in. In other words, it treats being logged in as an automatic opt-in for facet production.

Moreover, when the facet load strategy changes from any non-eager strategy to `'eager'`, any pending delayed facet fetch is supposed to go ahead immediately. However, currently no facet fetch is made in that scenario. This PR patches the bug to make that fetch.

### DIFF
--- a/src/collection-browser.ts
+++ b/src/collection-browser.ts
@@ -1411,14 +1411,19 @@ export class CollectionBrowser
    * current facet loading strategy.
    */
   private updateFacetReadiness(): void {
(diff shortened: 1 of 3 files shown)


iisa (approved):
👍

PR #391, opened 2024-08-13
WEBDEV-6918 Fix tile display for default sorted collections

latonv:
When viewing a collection that has a default sort set, the tiles are being sorted correctly but they fail to change the stats shown according to the default sort. For instance, when weekly views is the default (as it is for most collections), the tile stats still show the All-time Views rather than the Weekly Views amount. And when the default is a type of date, the creator line is not being replaced with a date line like it is supposed to.

This PR addresses some logic errors in how default sort params were being constructed & passed around, and ensures that the tiles display the correct info depending on the _effective_ sort (whether explicitly set or defaulted).

### DIFF
--- a/src/collection-browser.ts
+++ b/src/collection-browser.ts
@@ -122,8 +122,6 @@ export class CollectionBrowser
    */
   @property({ type: String }) displayMode?: CollectionDisplayMode;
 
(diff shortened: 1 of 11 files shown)


iisa (approved):
thanks for making it nice and clean on ingest 👍

PR #396, opened 2024-09-10
WEBDEV-6993 Fix logic error in tile overlays & improve code cleanliness

latonv:
There is currently a bug where items requiring login will show a "content may be inappropriate" overlay after logging in, even when they shouldn't. This PR fixes that bug and does some cleanup around the overlay components to make them clearer & more concise, and reduces a bunch of duplicated logic/templates across them. Also bolsters the tests to ensure the overlay logic is correct this time.


src/tiles/image-block.ts
```diff
@@ -51,35 +52,42 @@ export class ImageBlock extends LitElement {
...
 
-    return html`
-      <text-overlay
-        .loggedIn=${this.loggedIn}
-        .loginRequired=${this.model?.loginRequired}
-        ?iconRequired=${true}
-      >
-      </text-overlay>
-    `;
+    return html` <text-overlay .type=${this.overlayType}></text-overlay> `;
```
bfalling, line 77: QUESTION: The `?iconRequired` attribute is no longer needed?

bfalling (commented):
LGTM. Just one tiny question.

src/tiles/image-block.ts
latonv: Yes, this property was seemingly unused and perhaps a remnant of some earlier version of the component. Removed it entirely.

PR #402, opened 2024-09-26
WEBDEV-7022 Ensure initial page of results always triggers scroller refresh

latonv:
Broadens the conditions under which the infinite scroller's results will be refreshed, to always include the initial page of results (even when multiple pages are being loaded in simultaneously).


bfalling (approved):
LGTM. No apparent blockers. Code seems to make sense.

PR #403, opened 2024-10-01
Update CI to use newer codecov action

latonv:
Our existing means of uploading coverage reports to codecov has been deprecated and superseded by the codecov-action GH action. This PR updates our CI workflow to rely on the new action.


rebecca-shoptaw (approved):
LGTM! The failed test looks to be unrelated, but is something we should also fix at some point.

PR #417, opened 2025-01-15
Consolidated Item Removal process which can be used on offshoot pages.

nsharma123:
https://webarchive.jira.com/browse/WEBDEV-7228

### DIFF
--- a/package.json
+++ b/package.json
@@ -3,7 +3,7 @@
   "description": "The Internet Archive Collection Browser.",
   "license": "AGPL-3.0-only",
   "author": "Internet Archive",
-  "version": "2.7.6",
+  "version": "2.7.6-alpha2",
   "main": "dist/index.js",
   "module": "dist/index.js",
   "scripts": {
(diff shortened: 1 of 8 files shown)


src/manage/manage-bar.ts
```diff
@@ -32,11 +49,27 @@ export class ManageBar extends LitElement {
    */
   @property({ type: Boolean }) showUnselectAll = false;
 
+  /**
+   * Whether item has deleted or not (default true)
+   */
+  @property({ type: Boolean }) hasItemsDeleted = true;
```
latonv, line 55: I'm curious about how this is named & used. It seems incorrect that the default `true` implies items have already been deleted before anything happens. This is essentially just a switch for whether to show the error dialog, right?

Maybe this would be clearer in both ways if we inverted the logic, e.g., `hasRemoveItemsError = false;`

Alternatively, could we just call a method to pop up the error dialog when needed? I think it would be reasonable to handle opening the dialog via a method call rather than a reactive property, especially since there is no corresponding way to close the dialog via this component.

src/collection-browser.ts
```diff
@@ -220,6 +218,8 @@ export class CollectionBrowser
 
   @property({ type: String }) pageContext: CollectionBrowserContext = 'search';
 
+  @property({ type: String }) activeTabId: string = '';
+
```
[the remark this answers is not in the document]
nsharma123: Make sense. Thanks!

src/manage/manage-bar.ts
```diff
@@ -39,6 +27,11 @@ export class ManageBar extends LitElement {
    */
   @property({ type: Object }) selectedItems: Array<ManageableItem> = [];
 
+  /**
+   * Message shows as note in the modal when removing items
+   */
+  @property({ type: String }) manageViewModelMsg?: string;
```
latonv, line 33: Just noting a possible small typo: `Model` -> `Modal` ?

latonv (approved):
LGTM, nice work!

package.json
```diff
@@ -3,7 +3,7 @@
   "description": "The Internet Archive Collection Browser.",
   "license": "AGPL-3.0-only",
   "author": "Internet Archive",
-  "version": "2.7.6",
+  "version": "2.7.6-alpha4",
```
latonv, line 6: Reminder not to merge the alpha version into main.

PR #418, opened 2025-01-15
WEBDEV-7261 Fix discrepancy affecting old year-only publish dates

latonv:
Currently the date published shown on result tiles can be off by one for older years that fall before the creation of the local time zone, due to quirks in how dates are interpreted. This PR fixes the display error by ensuring that year-only dates always refer to UTC and do not involve the local time zone.

### DIFF
--- a/src/utils/format-date.ts
+++ b/src/utils/format-date.ts
@@ -20,8 +20,10 @@ export function formatDate(
 
   switch (format) {
     case 'year-only':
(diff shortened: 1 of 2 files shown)


rebecca-shoptaw (approved):
LGTM! Thanks for the explanatory comments and new tests!

latonv:
Addressing test failures in a separate ticket (they are unrelated to the current changes).

PR #419, opened 2025-01-17
WEBDEV-7371: Update ia-activity-indicator, histogram, and search-service

jbuckner:


rebecca-shoptaw:
Provisionally LGTM, but would you be willing to flag which files have the relevant edits we should be reviewing vs. just changed by the linter update? It's a bit tricky to comb through them, so maybe a list of files to check in the PR description or comments on the specific files?

jbuckner:
@rebecca-shoptaw The main changes here are updating the `package.json` file with dependency bumps. By updating search-service, I have to change the Metadata imports from `search-service` to `item-metadata`, then the TS WC template updates and linting. I don't believe there are any actual functional changes in here.

rebecca-shoptaw (approved):
LGTM!

PR #422, opened 2025-02-13
Large result for privs user when interact with manage interface

nsharma123:
https://webarchive.jira.com/browse/WEBDEV-7304

### DIFF
--- a/src/collection-browser.ts
+++ b/src/collection-browser.ts
@@ -263,10 +263,10 @@ export class CollectionBrowser
     new CollectionBrowserDataSource(this, this.pageSize);
 
   /**
-   * The maximum number of records that can be loaded for privileged users
-   * when clicking the "Manage" button on the search page.
+   * The maximum number of pages we will load when a privileged user clicks
+   * the "Manage" button on the search page. Limited to 15 pages.
    */
-  privsUserSearchThreshold = 750;
+  maxPagesToManage = 15;
 
   /**
    * The page that the consumer wants to load.
@@ -1459,7 +1459,7 @@ export class CollectionBrowser
     if (changed.has('isManageView')) {
       if (this.isManageView) {
         this.displayMode = 'grid';
-        this.fetchPrivsUserSearchResults();
+        this.fetchManagableSearchResults();
       }
 
       this.infiniteScroller?.refreshAllVisibleCells();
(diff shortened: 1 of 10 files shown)


src/data-source/collection-browser-data-source.ts
```diff
@@ -246,6 +246,19 @@ export class CollectionBrowserDataSource
     this.requestHostUpdate();
   }
 
+  /**
+   * @inheritdoc
+   */
+  resetPages(): void {
+    if (Object.keys(this.pages).length !== this.host.maxPagesToManage) {
+      this.pages = {};
```
latonv, line 254: Q: Do we need to reset `this.numTileModels` too since we are clearing them?

src/data-source/collection-browser-data-source.ts
```diff
@@ -246,6 +246,19 @@ export class CollectionBrowserDataSource
     this.requestHostUpdate();
   }
 
+  /**
+   * @inheritdoc
+   */
+  resetPages(): void {
+    if (Object.keys(this.pages).length !== this.host.maxPagesToManage) {
```
latonv, line 253: Q: This will still reset even if we already have more pages than needed. Would `<` instead of `!==` still work?

src/data-source/collection-browser-data-source.ts
```diff
@@ -246,6 +246,19 @@ export class CollectionBrowserDataSource
...
+      this.pages = {};
+    }
+
+    // Invalidate any fetches in progress
+    this.fetchesInProgress.clear();
+    this.requestHostUpdate();
```
latonv, line 259: Q: Should these lines be inside the `if` block too, or do we need them to run even if we don't reset?

src/collection-browser.ts
```diff
@@ -262,6 +262,12 @@ export class CollectionBrowser
   @property({ type: Object }) dataSource: CollectionBrowserDataSourceInterface =
     new CollectionBrowserDataSource(this, this.pageSize);
 
+  /**
+   * The maximum number of pages we will load when a privileged user clicks
+   * the "Manage" button on the search page. Limited to 15 pages.
+   */
+  maxPagesToManage = 15;
```
latonv, line 269: Q: Could we make this a `@property`?

src/collection-browser.ts
```diff
@@ -2086,6 +2097,24 @@ export class CollectionBrowser
...
+  /**
+   * Fetches search results for privileged users when in manage view
+   * If total results exceed the threshold, partially resets the datasource pages
+   * and fetches the first page with a limit based on the threshold
+   */
+  private fetchManagableSearchResults(): void {
+    const maxAllowedResults = this.maxPagesToManage * this.pageSize;
+    if (
+      this.pageContext === 'search' &&
+      this.dataSource.totalResults > maxAllowedResults &&
```
latonv, line 2109: Potential blocker: I'm not sure about the `this.dataSource.totalResults > maxAllowedResults` condition. That will prevent the larger fetch if there are, say, only 500 total results, right?

I think we can remove that condition, because we still want to fetch all the results in that case (and requesting 750 would still be fine)

latonv (commented):
Nice understandable approach! Noted one potential blocker for cases where there are < 750 total results, and a few questions that might need another look.

src/data-source/collection-browser-data-source.ts
nsharma123: We call the resetPages when needed. so inside/outside both work. I can just moving these inside. Thanks!

src/data-source/collection-browser-data-source.ts
nsharma123: We can but reset this create blank result placeholder while fetching 750 records. so just reset pages makes the  placeholder types to say, large search in-progress

src/collection-browser.ts
nsharma123: Make sense. I just need to change this logic a bit to make is more accurate.

src/collection-browser.ts
nsharma123: I thought about it but after that it won't be accessible in CollectionBrowserDataSource.ts

nsharma123: Yes, we can!

latonv (approved):
Thanks for quickly addressing the comments, LGTM!

PR #428, opened 2025-03-11
WEBDEV-7322 Ensure all facets can be scrolled into view

latonv:
Currently there are cases on some platforms where the facet bar flows slightly below the bottom of the viewport, making it impossible to scroll the last facet into view without scrolling the rest of the page too. This PR slightly adjusts how we size the facet sidebar to avoid that overflow situation.


src/collection-browser.ts
```diff
@@ -1612,10 +1612,12 @@ export class CollectionBrowser
       this.leftColumn?.style?.removeProperty('height');
     } else {
       const clientTop = this.leftColumn?.getBoundingClientRect().top;
-      this.leftColumn?.style?.setProperty(
-        'height',
-        `${window.innerHeight - (clientTop ?? 0) - 3}px`
-      );
+      const heightValue =
+        clientTop === 0
+          ? null
```
rebecca-shoptaw, line 1617: Much clearer with this separate variable, too

rebecca-shoptaw (approved):
LGTM!

src/collection-browser.ts
```diff
@@ -1612,10 +1612,12 @@ export class CollectionBrowser
...
+          : `${window.innerHeight - (clientTop ?? 0) - 20}px`;
```
bfalling, line 1618: SUGGESTION: Could comment on what this 20px is.

bfalling (approved):
One suggestion left. Otherwise, LGTM.

src/collection-browser.ts
latonv: I believe this is equivalent to the top padding of the facet container -- I think it would work to have this in a CSS var instead. Thanks!

PR #439, opened 2025-04-02
v2.8.0

latonv:
Version includes additional flexibility for enabling the management UI and controlling how many initial result pages are batched.


rebecca-shoptaw (approved):
Beautiful

PR #447, opened 2025-05-23
WEBDEV-7479 Tile updates for TV collections/clips

latonv:
Some minor tile changes are required to display results within TV collections, which may be "clips" of a larger program. Tiles in these collections should also display how many times they have been "clipped" in place of their review count.

We also need slight adjustments to the data source class and main component to recognize TV collections and set the correct state when they are encountered.


src/tiles/grid/tile-stats.ts
```diff
@@ -1,88 +1,142 @@
...
-            </p>
-          </li>
+          ${this.mediatypeIconColumnTemplate}
+          ${this.mediatype === 'account'
+            ? this.uploadsColumnTemplate
+            : this.viewsColumnTemplate}
+          ${this.favoritesColumnTemplate}
+          ${this.showTvClips
+            ? this.tvClipsColumnTemplate
+            : this.reviewsColumnTemplate}
```
latonv, line 54: This render method was getting somewhat unwieldy given all the mixing of conditional logic and HTML throughout it. I refactored things to pull out the individual column templates and make the logic more readable.

src/data-source/collection-browser-data-source.ts
```diff
@@ -1132,6 +1138,14 @@ export class CollectionBrowserDataSource
         this.parentCollections = [].concat(
           this.collectionExtraInfo.public_metadata?.collection ?? [],
         );
+
+        // Update the TV collection status now that we know the parent collections
+        this.host.isTVCollection =
+          this.host.withinCollection?.startsWith('TV-') ||
```
bfalling, line 1144: QUESTION: We're okay with this being case-sensitive?

src/tiles/grid/tile-stats.ts
```diff
@@ -1,88 +1,142 @@
...
+   * @param classes Any additional CSS classes the stat column should have (optional).
+   */
+  private columnTemplate(
+    count: number | undefined,
+    label: string,
+    icon: TemplateResult,
+    classes: string[] = [],
+  ): TemplateResult {
+    const formattedCount = formatCount(count ?? 0, 'short', 'short');
+    const title = `${formattedCount} ${label}`;
```
bfalling, line 88: QUESTION: Does a title like this need to go through `msg()`?

PR #461, opened 2025-07-04
WEBDEV-7610 Check default sort applied by PPS instead of metadata

latonv:
Currently to determine the default sort option to select for a collection, we check for its `sort-by` metadata and fall back to global defaults otherwise. However, TV collections have their own default that is not always specified in metadata. And since we can't always identify TV collections until after receiving a PPS response, we need to rely on the PPS layer to both apply and inform the client of the correct default sort.

This PR updates how we determine which sort option to select, relying on the PPS's finalized request parameters rather than the collection metadata for this purpose.

### DIFF
--- a/src/models.ts
+++ b/src/models.ts
@@ -449,7 +449,7 @@ export const SORT_OPTIONS: Record<SortField, SortOption> = {
     handledBySearchService: false,
     searchServiceKey: 'favoritedate',
     displayName: 'Date favorited',
-    urlNames: ['favoritedate'],
+    urlNames: ['favoritedate', 'date_favorited'],
   },
   [SortField.creator]: {
     field: SortField.creator,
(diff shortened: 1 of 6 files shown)


src/models.ts
```diff
@@ -449,7 +449,7 @@ export const SORT_OPTIONS: Record<SortField, SortOption> = {
     handledBySearchService: false,
     searchServiceKey: 'favoritedate',
     displayName: 'Date favorited',
-    urlNames: ['favoritedate'],
+    urlNames: ['favoritedate', 'date_favorited'],
```
latonv, line 452: The PPS identifies this sort using `date_favorited`, a format that differs from the other date sorts because of the different backing store for favorites & date favorited. Adding the new one so that we recognize it correctly, and leaving the old one for backwards-compatibility.

PR #462, opened 2025-07-09
WEBDEV-7452 Implement TV tiles

latonv:
Updates the tile models & display components to correctly display special mediatype icons for TV items that are part of certain collections.


src/data-source/collection-browser-data-source.ts
```diff
@@ -1220,7 +1221,14 @@ export class CollectionBrowserDataSource
...
     results?.forEach(result => {
       if (!result.identifier) return;
-      tiles.push(new TileModel(result));
+      const requestSource: HitRequestSource = this.host.baseQuery
+        ? 'search_query'
+        : this.host.withinProfile
+          ? 'profile_tab'
+          : this.host.withinCollection
+            ? 'collection_members'
+            : 'unknown';
```
jbuckner, line 1230: Suggestion: pull this ternary out to a getter to help improve the readability of this `forEach()` (and maybe make it a series of `if`s instead of a ternary, also for readability). Also, since it's a one-time computed value, you can move it outside of the `forEach` so it doesn't have to compute it each time. ie:
```
private get hitRequestSource(): HitRequestSource {
  if (this.host.baseQuery) return 'search_query';
  if (this.host.withinProfile) return 'profile_tab';
  ...
}
...
const requestSource = this.hitRequestSource;
forEach( ...
  new TileModel(result, requestSource)
)
```

latonv: Done, thanks for catching that!

src/tiles/mediatype-icon.ts
```diff
@@ -1,33 +1,69 @@
...
+  mediatypeConfig,
+  MediatypeConfigKey,
+} from '../mediatype/mediatype-config';
+
+const TV_COMMERCIAL_COLLECTION = 'tv_ads';
+const TV_FACT_CHECK_COLLECTION = 'factchecked';
+const TV_COLLECTIONS = new Set(['tvnews', 'tvarchive', 'television']);
+const RADIO_COLLECTIONS = new Set(['radio', 'radioprogram']);
 
 @customElement('mediatype-icon')
```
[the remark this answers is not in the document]
latonv: Yeah, good point. Perhaps `tile-mediatype-icon` could work here, to keep the name more compact and specific to the result tiles that it's designed for?

PR #487, opened 2025-09-09
WEBDEV-7713 Upgrade date picker to fix tooltip clipping

latonv:
The date picker's tooltips currently get clipped by the sides of the facet column, impacting readability of tooltips for bars near the start and end of the histogram. This PR upgrades the date picker to a bugfixed version that uses popovers to avoid clipping.

### DIFF
--- a/package.json
+++ b/package.json
@@ -26,7 +26,7 @@
     "@internetarchive/analytics-manager": "^0.1.4",
     "@internetarchive/feature-feedback": "^1.0.0",
     "@internetarchive/field-parsers": "^1.0.0",
-    "@internetarchive/histogram-date-range": "^1.3.1",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -26,7 +26,7 @@
     "@internetarchive/analytics-manager": "^0.1.4",
     "@internetarchive/feature-feedback": "^1.0.0",
     "@internetarchive/field-parsers": "^1.0.0",
-    "@internetarchive/histogram-date-range": "^1.3.1",
+    "@internetarchive/histogram-date-range": "1.3.2-alpha-webdev7713.0",
```
latonv, line 29: Will bump to a non-alpha version before merging.

PR #494, opened 2025-09-24
WEBDEV-7757 Adjust sort options shown for TV searches

latonv:
TV searches do not require all the same sort options available, in particular since many of the date options are either not relevant or essentially the same for TV items.

This PR generalizes how the sort bar component is configured a bit, so that we can easily modify what sort options are available in bulk rather than via individual flags, and then ensures it is configured appropriately in the cases that require non-default sort options.

### DIFF
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -4,7 +4,7 @@
     "module": "esnext",
     "moduleResolution": "node",
     "noEmitOnError": true,
-    "lib": ["es2017", "dom"],
+    "lib": ["es2017", "dom", "dom.iterable"],
     "strict": true,
     "esModuleInterop": false,
     "allowSyntheticDefaultImports": true,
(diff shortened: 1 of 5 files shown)


tsconfig.json
```diff
@@ -4,7 +4,7 @@
     "module": "esnext",
     "moduleResolution": "node",
     "noEmitOnError": true,
-    "lib": ["es2017", "dom"],
+    "lib": ["es2017", "dom", "dom.iterable"],
```
iisa, line 7: cool! should we also add this to the main ts template?

iisa (approved):

tsconfig.json
latonv: Probably, yeah. I'll create an MR to add it there too.

PR #558, opened 2026-02-25
WEBDEV-8186 Implement new single-dropdown sort bar

latonv:
Currently the layout of the sort bar at large widths is to split out several different groups of sort options, only collapsing them to a single dropdown below a mobile breakpoint. However, for improved clarity and consistency, we will be switching to a single-dropdown approach in all cases.

This PR removes the old grouped desktop layout and renders the single-dropdown layout for all cases (as in the old mobile version). Moreover, the design of the sort direction toggle is adjusted to make it clearer that it is interactive.


bfalling (approved):
LGTM

PR #565, opened 2026-04-22
Favorites Tab: Update collection-browser to support `favorites` tab

nsharma123:
https://webarchive.jira.com/browse/WEBDEV-8351

Description
- Added support for the favorites tab in collection-browser
- Mapped collection_extra_info to render favorites list
- Updated integration to latest search-service
- Minor refactor for cleaner tab-specific logic


package.json
```diff
@@ -30,10 +30,11 @@
     "@internetarchive/field-parsers": "^1.0.0",
     "@internetarchive/histogram-date-range": "^1.4.1",
     "@internetarchive/ia-dropdown": "^2.0.0",
+    "@internetarchive/ia-userlist-settings": "^1.1.3-alpha3",
```
latonv: We shouldn't need this dependency here in collection-browser. Can we remove?

package.json
```diff
@@ -3,7 +3,7 @@
   "description": "The Internet Archive Collection Browser.",
   "license": "AGPL-3.0-only",
   "author": "Internet Archive",
-  "version": "4.1.1",
+  "version": "4.1.2-alpha12",
```
latonv: (Just a reminder to adjust this before merging)

package.json
nsharma123: Oh yeah, it should not be there. Thanks for catching this

src/data-source/collection-browser-data-source.ts
```diff
@@ -1127,7 +1127,7 @@ export class CollectionBrowserDataSource
       this.host.defaultSortField
     ) {
       const sortOption = SORT_OPTIONS[this.host.defaultSortField];
-      if (sortOption.searchServiceKey) {
+      if (sortOption.searchServiceKey && sortOption.handledBySearchService) {
```
iisa: can we use destructure? const { searchKeyService, handledBySearchService } = sortOption;

iisa (approved):
looks good, just one comment. also, would be good to add unit test

PR #576, opened 2026-05-07
Collection Browser – Add delay note to favorites removal modal

nsharma123:
https://webarchive.jira.com/browse/WEBDEV-8432

### DIFF
--- a/src/app-root.ts
+++ b/src/app-root.ts
@@ -471,7 +471,8 @@ export class AppRoot extends LitElement {
             </div>
             <details class="profile-element-controls">
               <summary>
-                Profile tab (profileElement)${this.profileElement
+                Profile tab
+                (profileElement)${this.profileElement
                   ? html`: <strong>${this.profileElement}</strong>`
                   : ''}
               </summary>
--- a/src/app-root.ts
+++ b/src/app-root.ts
@@ -53,6 +53,8 @@ export class AppRoot extends LitElement {
 
   @state() private profileElement?: PageElementName;
 
+  @state() private withinProfile?: string;
+
   @property({ type: Object, reflect: false }) latestAction?: AnalyticsEvent;
 
   @query('#base-query-field') private baseQueryField!: HTMLInputElement;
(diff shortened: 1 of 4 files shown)


src/app-root.ts
```diff
@@ -466,6 +469,51 @@ export class AppRoot extends LitElement {
                 Show replaced sort options
               </label>
             </div>
+            <details class="profile-element-controls">
+              <summary>
+                Profile tab
+                (profileElement)${this.profileElement
+                  ? html`: <strong>${this.profileElement}</strong>`
+                  : ''}
+              </summary>
```
latonv, line 492: Thanks for adding these controls to the demo app! It also occurs to me that the demo still lacks a way to specify the value of `withinProfile`, which would be set to a user item identifier like `@foobar`. Should we maybe add that too, to make these new `profileElement` controls more meaningful?

src/manage/manage-bar.ts
```diff
@@ -105,6 +106,34 @@ export class ManageBar extends LitElement {
...
+      case 'favorites':
+        listName = 'favorites list';
+        break;
+      default:
+        return '';
+    }
+
+    return msg(
+      str`Note: It may take a few minutes for ${subject} to stop appearing in your ${listName}.`,
+    );
```
latonv, line 134: Nice simplification/DRY of this getter, the logic reads much clearer now.

test/manage/manage-bar.test.ts
```diff
@@ -145,13 +148,200 @@ describe('Manage bar', () => {
...
+      .exist;
+  });
+
+  it('emits manageItems event when Item Manager button clicked', async () => {
+    const spy = Sinon.spy();
+    const el = await fixture<ManageBar>(html`
+      <manage-bar
+        showItemManageButton
+        removeAllowed
+        @manageItems=${() => spy()}
```
latonv: Couldn't this just be `@manageItems=${spy}`? Or does tsc balk at that?

(Similar question for some of the other tests)

latonv (approved):
LGTM, just had a few small suggestions.

src/app-root.ts
nsharma123: Good idea!. thanks!

test/manage/manage-bar.test.ts
nsharma123: Yes, I could be. Changed!

PR #577, opened 2026-05-07
WEBDEV-7004 Ensure collections fall back to metadata search for empty queries, even with FTS selected

latonv:
The FTS engine does not handle empty queries, even within collections. So when a patron clears their FTS query on a collection, we currently (incorrectly) show an error indicating there are no viewable items. In such cases, we actually just want to show the full collection again, via metadata search. This PR ensures that we fall back to metadata in that specific case of "in collection + empty query + FTS selected".

### DIFF
--- a/src/data-source/collection-browser-data-source.ts
+++ b/src/data-source/collection-browser-data-source.ts
@@ -604,6 +604,24 @@ export class CollectionBrowserDataSource
   }
 
   /**
+   * Returns the search type to actually use for fetches.
+   * When within a collection with an empty query and FTS selected,
+   * falls back to metadata search since FTS requires a non-empty query
+   * and we'd still like to show the unfiltered collection in that case.
+   */
+  private get effectiveSearchType(): SearchType {
+    const trimmedQuery = this.host.baseQuery?.trim();
+    if (
+      !trimmedQuery &&
+      this.host.withinCollection &&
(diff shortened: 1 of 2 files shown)


src/data-source/collection-browser-data-source.ts
```diff
@@ -603,6 +603,24 @@ export class CollectionBrowserDataSource
...
+  /**
+   * Returns the search type to actually use for fetches.
+   * When within a collection with an empty query and FTS selected,
+   * falls back to metadata search since FTS requires a non-empty query
+   * and we'd still like to show the unfiltered collection in that case.
+   */
+  private get effectiveSearchType(): SearchType {
+    const trimmedQuery = this.host.baseQuery?.trim();
+    if (
+      !trimmedQuery &&
```
rebecca-shoptaw, line 616: Super minor but I think an intermediate variable could be helpful here, i.e. `shouldUseMetadataSearch = !trimmedQuery && ...` so that you don't have to put the whole conditional in the if statement and it's clear why you're checking is someone didn't read the comment

rebecca-shoptaw (approved):
An elegant solution, and thanks for adding the tests! One minor clarity note, but looks great.

src/data-source/collection-browser-data-source.ts
latonv: Makes sense, updated!

PR #578, opened 2026-05-09
WEBDEV-7901 Indicate "Weekly" or "All-time" views on tile labels

latonv:
On our list and compact-list tiles, we currently update the number of views according to the current sort option. However, on our list and compact-list tiles, this is currently still just labeled "Views" with no indication of which type of views it represents. This PR updates those labels to more precisely indicate whether they represent "Weekly views" or "All-time views".

### DIFF
--- a/src/tiles/list/tile-list.ts
+++ b/src/tiles/list/tile-list.ts
@@ -250,10 +250,10 @@ export class TileList extends BaseTileComponent {
   }
 
   private get viewsTemplate() {
-    const viewCount =
-      this.effectiveSort?.field === 'week'
-        ? this.model?.weeklyViewCount // weekly views
-        : this.model?.viewCount; // all-time views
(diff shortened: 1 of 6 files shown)


src/tiles/list/tile-list.ts
```diff
@@ -250,10 +250,10 @@ export class TileList extends BaseTileComponent {
   }
 
   private get viewsTemplate() {
-    const viewCount =
-      this.effectiveSort?.field === 'week'
-        ? this.model?.weeklyViewCount // weekly views
-        : this.model?.viewCount; // all-time views
+    const isWeeklySort = this.effectiveSort?.field === 'week';
+    const viewCount = isWeeklySort
```
rebecca-shoptaw, line 254: Thanks for doing this cleanup while you were here!

rebecca-shoptaw (approved):
LGTM!

iisa (approved):

PR #580, opened 2026-05-22
WEBDEV-8495 Upgrade date picker to fix regression in Lit 3

latonv:
When the version of Lit in use is overridden to Lit 3, the `histogram-date-range` component currently has a bug where it does not emit any events when the date range is updated. This PR just upgrades that component to a fixed version.

### DIFF
--- a/package.json
+++ b/package.json
@@ -28,7 +28,7 @@
     "@internetarchive/elements": "^0.2.2",
     "@internetarchive/feature-feedback": "^1.0.0",
     "@internetarchive/field-parsers": "^1.0.0",
-    "@internetarchive/histogram-date-range": "^1.4.1",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -28,7 +28,7 @@
     "@internetarchive/elements": "^0.2.2",
     "@internetarchive/feature-feedback": "^1.0.0",
     "@internetarchive/field-parsers": "^1.0.0",
-    "@internetarchive/histogram-date-range": "^1.4.1",
+    "@internetarchive/histogram-date-range": "1.4.2-alpha-webdev8495.0",
```
latonv: (To be upgraded to an equivalent non-alpha version before merging)

PR #584, opened 2026-05-29
WEBDEV-8334 Fix percent-encoded URLs for web archives capture links

dualcnhq:
## Changes
- Remove encodeURIComponent() from the capture href construction
- Decode percent-encoded href values in TileDispatcher to fix broken web archive capture links
- Update tests for percent-encoded href fixes

### DIFF
--- a/package.json
+++ b/package.json
@@ -3,7 +3,7 @@
   "description": "The Internet Archive Collection Browser.",
   "license": "AGPL-3.0-only",
   "author": "Internet Archive",
-  "version": "4.3.2-rc-webdev-8334.4",
+  "version": "4.3.2-rc-webdev-8334.5",
   "main": "dist/index.js",
   "module": "dist/index.js",
   "scripts": {
@@ -33,7 +33,7 @@
     "@internetarchive/iaux-item-metadata": "^1.0.5",
     "@internetarchive/infinite-scroller": "^2.0.0",
     "@internetarchive/modal-manager": "^2.0.5",
-    "@internetarchive/search-service": "2.7.2-rc-webdev-8334.2",
+    "@internetarchive/search-service": "2.7.2",
     "@internetarchive/shared-resize-observer": "^0.2.0",
     "@lit/localize": "^0.12.2",
     "dompurify": "^3.2.4",
--- a/package.json
+++ b/package.json
@@ -3,7 +3,7 @@
   "description": "The Internet Archive Collection Browser.",
   "license": "AGPL-3.0-only",
   "author": "Internet Archive",
-  "version": "4.3.2-rc-webdev-8334.0",
+  "version": "4.3.2-rc-webdev-8334.4",
   "main": "dist/index.js",
   "module": "dist/index.js",
   "scripts": {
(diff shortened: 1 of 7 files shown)


src/tiles/tile-dispatcher.ts
```diff
@@ -176,7 +176,12 @@ export class TileDispatcher
     // Use the server-specified href if available.
     // Otherwise, construct a details page URL from the item identifier.
     if (this.model.href) {
-      return `${this.baseNavigationUrl}${this.model.href}`;
+      // Defensive decode: %3A (encoded colon) in a URL is an unambiguous sign the
+      // target URL was incorrectly percent-encoded — decode it before use.
+      const href = /%3A/i.test(this.model.href)
+        ? decodeURIComponent(this.model.href)
+        : this.model.href;
+      return `${this.baseNavigationUrl}${href}`;
```
latonv: I think this signal may be more ambiguous than implied. Consider a full-text search for a query like "animals: cats & dogs". The result tiles for that query all include `?q=%22animals%3A+cats+%26+dogs%22` in their `href` to propagate the query into the item itself. If we decode those `href` values, everything after the `&` gets dropped from the query and treated as a separate URL param.

Suggestion: could we gate this decoding on the existence of `model.captureDates`, to ensure we only apply it to web archive hits and not other cases?

latonv: Revisiting this -- can this extra logic be removed now that the new search service version no longer encodes the href value in the first place?

src/tiles/tile-display-value-provider.ts
```diff
@@ -124,9 +124,8 @@ export class TileDisplayValueProvider {
       .toISOString()
       .replace(/[TZ:-]/g, '')
       .replace(/\..*/, '');
-    const captureHref = `https://web.archive.org/web/${captureDateStr}/${encodeURIComponent(
-      url,
-    )}`;
+    // url must not be percent-encoded — Wayback Machine matches on the raw URL
+    const captureHref = `https://web.archive.org/web/${captureDateStr}/${url}`;
```
latonv, line 128: 👍

package.json
```diff
@@ -3,7 +3,7 @@
   "description": "The Internet Archive Collection Browser.",
   "license": "AGPL-3.0-only",
   "author": "Internet Archive",
-  "version": "4.4.0",
+  "version": "4.3.2-rc-webdev-8334.4",
```
latonv: (Just flagging this so we don't merge this version string)

src/tiles/tile-display-value-provider.ts
dualcnhq: thanks for the review

src/tiles/tile-dispatcher.ts
dualcnhq: reverted

package.json
dualcnhq: reverted as well

latonv (approved):
LGTM, thanks for the fixes!

